PR #76, opened 2026-05-11
fix(projects): track shell session liveness via PTY output recency

im-ian:
## Summary
Follow-up to #73. Shell sessions never produce a Claude Code transcript at `~/.claude/projects/*/<id>.jsonl`, so the existing status detector had no signal for them and forever reported `Idle`. The Sidebar status dot stayed grey for plain `$SHELL` sessions even while the user was actively typing or running commands.

This wires a PTY output timestamp into the detector:

- `PtyHandle` now holds an `Arc<Mutex<Instant>>` initialized at spawn.
- The PTY reader thread updates it on every non-empty `read()`.
- `PtyManager::last_output_at(session_id)` exposes the latest instant.
- `session_status::detect` takes `Option<Instant>` and falls back to PTY recency when no Claude transcript exists — `< 2500 ms` since last output → `Running`, otherwise `Idle`.
- Claude (Agent) sessions still use the transcript walker → finer-grained `NeedsInput` vs `Running` classification.
- `commands::detect_session_statuses` threads `state.pty.last_output_at(uuid)` through.

The 2500 ms window covers two 1 s frontend polls, so `Running` stays stable for at least 2 ticks after each output burst before flipping back to `Idle`. Long commands with sparse output may briefly oscillate; that's intentional — the dot reflects "did anything happen recently" not "is a process alive".

## Test plan
- [x] `cargo test --lib session_status` 12 pass (3 new for PTY fallback)
- [x] `cargo check` clean
- [x] `bun run build` clean
- [x] `bun run test` 87 pass / 1 skipped
- [ ] Live: open a shell session, dot is `Idle` while quiet
- [ ] Live: type / run a command in shell, dot flips to `Running`
- [ ] Live: dot returns to `Idle` ~2.5 s after output stops
- [ ] Live: Claude Agent session dot still reflects transcript-based status (`NeedsInput`/`Running`)

### DIFF
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -629,7 +629,11 @@ pub async fn detect_session_statuses(
                 .and_then(|uuid| state.sessions.get(&uuid).ok())
                 .map(|s| s.status)
                 .unwrap_or(SessionStatus::Idle);
-            let status = session_status::detect(&id, previous).unwrap_or(previous);
+            // Shell sessions have no Claude transcript — fall back to PTY
+            // output recency for liveness. `None` = no live PTY → Idle.
+            let last_pty_output = parsed_id.and_then(|uuid| state.pty.last_output_at(&uuid));
+            let status = session_status::detect(&id, previous, last_pty_output)
+                .unwrap_or(previous);
             // Mirror the detected status into the in-memory store so persisted
             // sessions reflect liveness on next save. Best-effort: ignore errors
             // (e.g. UUID parse failure for a stale id from the frontend).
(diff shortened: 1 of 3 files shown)


im-ian:
Closing in favor of a more accurate approach: track child-process transitions instead of raw PTY output recency.

The PTY-output timestamp here flips a shell session to `Running` whenever the user types (because the shell echoes the keystroke), which conflates "user is typing" with "command is executing." It also can't surface a `NeedsInput` signal when a long-running command finishes — a feature parity request from the agent-mode `NeedsInput` dot.

Follow-up branch will:
- Track each shell PID's child processes via `sysinfo`.
- Map child presence → `Running`, child-exit transition → sticky `NeedsInput`, quiet → `Idle`.
- Clear sticky `NeedsInput` on the next PTY write (user moved on).

Will link the new PR here when up.

PR #92, opened 2026-05-11
fix(terminal): repaint viewport after PTY output idle

im-ian:
## Summary

Force a viewport repaint shortly after a PTY output burst goes idle, so xterm's DOM renderer doesn't leave stale glyphs visible after an interrupted TUI redraw.

## Why

xterm's DOM renderer reuses per-row DOM elements and incrementally patches glyphs as the buffer changes. When a streaming TUI (Claude CLI, `htop`) is interrupted mid-redraw — user hits Esc/Ctrl+C, or the stream simply stops — the parser settles but the last cell-diff set may never trigger a paint that overwrites stale glyphs left by an earlier wider line. The xterm buffer is correct (copy-to-clipboard yields clean text); only the DOM rendition is stale, so the user sees garbled overlapping characters.

We can't drop the DOM renderer (canvas/webgl break CJK IME positioning), so we work around the renderer instead: `term.refresh(0, term.rows - 1)` rebuilds every visible row directly from the buffer, clearing leftover characters.

## How

- Add a `scheduleViewportRepaint()` debounced timer, fired from the existing `pty:output` listener alongside `scheduleScrollbackSave()`.
- Idle threshold: 120ms after the last chunk. Short enough that the stale frame is rarely visible, long enough to coalesce a streaming burst into a single refresh.
- Cleanup clears the pending timer on unmount.

Mid-burst paints already happen naturally as each chunk lands, so we only fire when the stream goes quiet — no per-chunk cost.

## Test plan

- [x] `bun run typecheck`
- [x] `bun run test` (89 passed)
- [ ] Manual: run `claude` in a session, interrupt mid-stream with Esc / Ctrl+C, confirm no stale overlapping glyphs remain
- [ ] Manual: run `htop`, resize panel during refresh, confirm rows redraw cleanly
- [ ] Manual: confirm CJK IME composition preview still positions correctly after the repaint fires

### DIFF
--- a/src/components/Terminal.tsx
+++ b/src/components/Terminal.tsx
@@ -199,6 +199,42 @@ export function Terminal({
       }, SCROLLBACK_SAVE_DEBOUNCE_MS);
     };
 
+    // Force a viewport repaint shortly after a PTY output burst goes idle.
+    //
+    // xterm's DOM renderer reuses per-row DOM elements and incrementally
+    // patches glyphs as the buffer changes. When a streaming TUI (Claude
+    // CLI, `htop`, `claude --print`) is interrupted mid-redraw — user
+    // presses Esc/Ctrl+C, or the stream simply stops — the parser settles
+    // but the last set of cell-diffs may never trigger a paint that
+    // overwrites stale glyphs left by an earlier wider line. The xterm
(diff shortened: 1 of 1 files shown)


im-ian:
Superseded by #93 (branch renamed to `fix/terminal-stale-glyph-repaint`).

PR #156, opened 2026-05-13
feat(sessions): auto-show worktree icon when session lives inside a worktree

im-ian:
## Summary

Auto-show the worktree icon on session tabs and sidebar rows whenever the session lives inside a linked git worktree. Before, only Acorn's "new isolated session" button (`isolated=true`) lit the icon. Now `claude -w` adoptions, projects added at a worktree path, and interactive `cd` into a worktree all flag automatically.

## How it works

Three signals combined by precedence:

- `Session.isolated` — Acorn ownership (creation-time fact, unchanged removal semantics).
- `Session.in_worktree` — derived backend-side on every `list_sessions` by stat'ing `worktree_path/.git`. `skip_deserializing` so it never persists stale. Reflects `claude -w` adoptions as soon as the new worktree_path is recorded.
- `liveInWorktree[id]` — live PTY cwd walked through descendants. Refreshed event-driven only (after `refreshSessions` and on window focus). No interval polling.

Icon condition: `liveInWorktree ?? (isolated || in_worktree)`. When a live PTY exists, the live signal wins — `cd` out of a worktree clears the icon; `cd` into one lights it on the next focus. Without a live PTY the static flags apply as fallback.

## Cost

- `pty_in_worktree_all` does one `System::refresh_processes_specifics(All, with_cwd)` + a BFS per session. ~20-30ms total regardless of session count, vs. that cost × N when looping `pty_cwd` per session.
- Event-driven invocation only (window focus + post-refresh). No interval polling.

## Test plan

- [x] `bun run typecheck`
- [x] `bun run test` (123 passed)
- [x] `cargo check`
- [x] `cargo clippy` (new code clean; remaining warnings pre-existing)
- [ ] Acorn isolated session → icon shows on tab + sidebar
- [ ] `claude -w` creates worktree + adopts → icon shows immediately after adoption
- [ ] Inside a worktree, run `cd /tmp` → icon disappears on next focus
- [ ] Outside a worktree, run `cd <linked-worktree>` → icon appears on next focus
- [ ] `.acorn/worktrees/<name>` worktrees still flag (regression check)

### DIFF
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -171,13 +171,20 @@ pub fn list_sessions(state: State<'_, AppState>) -> Vec<Session> {
         .sessions
         .list()
         .into_iter()
-        .map(|mut s| {
-            if let Ok(branch) = worktree::current_branch(&s.worktree_path) {
-                s.branch = branch;
-            }
-            s
-        })
+        .map(enrich_session)
         .collect()
+}
+
+/// Attach derived fields (`branch`, `in_worktree`) computed from the
+/// session's current on-disk state. Pure: never mutates the persisted
+/// store. Called on every Session leaving the backend so the frontend
+/// sees fresh values without a second round-trip.
+fn enrich_session(mut s: Session) -> Session {
+    if let Ok(branch) = worktree::current_branch(&s.worktree_path) {
+        s.branch = branch;
+    }
+    s.in_worktree = worktree::is_linked_worktree_root(&s.worktree_path);
+    s
 }
 
 static MEMORY_PROBE: Mutex<Option<System>> = Mutex::new(None);
@@ -320,7 +327,7 @@ pub async fn create_session(
     let inserted = state.sessions.insert(session);
     state.projects.ensure(repo.clone(), project_basename(&repo));
     persist(&state);
-    Ok(inserted)
+    Ok(enrich_session(inserted))
 }
 
 #[tauri::command]
@@ -360,7 +367,7 @@ pub fn reorder_sessions(
         .collect();
     state.sessions.reorder(&path, &ids);
     persist(&state);
-    Ok(state.sessions.list())
+    Ok(state.sessions.list().into_iter().map(enrich_session).collect())
 }
 
 #[tauri::command]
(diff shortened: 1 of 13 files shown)


im-ian (commented):
## Self-review

**Decision**: APPROVE — no critical/high findings.

### Validation

- `bun run typecheck` pass
- `bun run test` pass (123 / 124, 1 skipped)
- `cargo check` pass
- `cargo clippy` pass (new code clean after fffd81e; remaining warnings pre-existing)

### Medium

**Tests missing for new behavior.** `is_linked_worktree_root` + `pty_in_worktree_all` + `store.refreshLiveInWorktree` have no coverage. The `live ?? static` precedence in `Pane.tsx` / `Sidebar.tsx` is exactly the rule that regresses silently. Adding a Rust unit test for `is_linked_worktree_root` (temp dir + `Repository::init` + `worktree`) and a `store.test.ts` case for `refreshLiveInWorktree` populating + falling back would close the gap cheaply.

### Low

1. `SessionRowPreview` (Sidebar.tsx:462) drag overlay uses static-only condition (`isolated || in_worktree`). Visible only mid-drag, so kept hook-free.
2. `enrich_session` is opt-in per command — future commands returning `Session` could forget. Could push into `SessionStore` API but mixes derived state into the store contract.
3. Window `focus` could fire often if user alt-tabs rapidly; each refire = one ~20-30ms backend call. Acceptable at typical rates; debounce only if observed.
4. Brief inconsistency window if PTY child dies before its handle is evicted from the registry — `pty_in_worktree_all` returns `false` instead of falling through to static. Self-corrects on next refresh.

### Pre-merge checklist (from PR body)

`bun` test plan still has interactive items unchecked:

- Acorn isolated session → icon
- `claude -w` adoption → icon
- `cd` out of worktree → icon disappears next focus
- `cd` into worktree → icon appears next focus
- `.acorn/worktrees/<name>` regression

These need a live build to verify — port 1430 was in use during review so the dev server was not started. Per the project memory rule (`feedback_pr_body_sync_before_merge.md`), check these off in the PR body once verified.

Full report: `.claude/PRPs/reviews/pr-156-review.md`

PR #158, opened 2026-05-13
feat(sessions): emit OSC 7 from zsh + xterm handler for instant cwd-based worktree icon

im-ian:
## Summary

Make the worktree icon update **instantly** on `cd` without polling. Builds on #156, which used a focus-event refresh + a 20-30ms backend process scan per call. This PR replaces the polling path for zsh sessions with shell-side OSC 7 emission: zero CPU when idle, real-time on every prompt.

Stacks on top of #156 — merge that first.

## How it works

Three pieces:

1. **Bundled zsh rc** (`src-tauri/shell-init/zshrc`, `include_str!` into the binary). Materialised under the data dir on every PTY spawn, mirrors the `agent_shim` pattern. Restores the user's original `ZDOTDIR`, sources their real `.zshrc` so their config still runs, then registers a `precmd` hook that emits `\e]7;file://<host><pwd>\e\\` after every command.

2. **PTY env injection** — `pty_spawn` sets `ZDOTDIR` to Acorn's staged dir and stashes the user's original under `ACORN_USER_ZDOTDIR`. `.zshenv` from `$HOME` still loads unconditionally, so config that has to land before any rc is untouched.

3. **xterm OSC 7 handler** + `is_path_linked_worktree(path)` invoke. Each shell emit hands the host a fresh cwd; the backend classifies it via libgit2 `Repository::discover` + `.git` is-file check; the result lands in `liveInWorktree[id]`. No process-table refresh, no polling.

## Scope

zsh only. macOS default + ~95% of macOS sessions covered. bash supports `PROMPT_COMMAND` from env directly (later); fish emits OSC 7 by default (no work needed). The #156 focus-event refresh stays as the fallback for non-zsh shells.

## Test plan

- [x] `bun run typecheck`
- [x] `bun run test` (123 passed)
- [x] `cargo clippy` (clean for new code)
- [x] `cargo test shell_init` (2 passed)
- [x] Live verified — new PTY emits OSC 7 every prompt, backend classifies, `cd` into a worktree updates the icon immediately, `cd` out clears it (confirmed via webview console + UI)
- [ ] User's existing `~/.zshrc` still runs (sanity check on a non-trivial config)
- [ ] `.acorn/worktrees/<name>` and external linked worktrees (`claude -w`) both light the icon

### DIFF
--- a/src-tauri/shell-init/zshrc
+++ b/src-tauri/shell-init/zshrc
@@ -1,0 +1,32 @@
+# Acorn zsh init.
+#
+# Sourced because Acorn sets `ZDOTDIR` to its own staged dir before spawning
+# the PTY. This file restores the user's real `ZDOTDIR`, sources their
+# `.zshrc` so their configuration still runs, then appends an OSC 7
+# emitter so the terminal learns the live cwd after every command without
+# the host having to poll `proc_pidinfo`.
+#
+# OSC 7 (`\e]7;file://<host><pwd>\e\\`) is the de-facto "current directory"
+# protocol shared by iTerm2, GNOME Terminal, Wezterm, Kitty, VS Code, etc.
+# Acorn's xterm.js handler treats every emit as the new live cwd for the
+# session and updates the worktree-icon condition in O(1) — no system scan.
+
+if [ -n "${ACORN_USER_ZDOTDIR-}" ]; then
+    ZDOTDIR=$ACORN_USER_ZDOTDIR
+else
+    ZDOTDIR=$HOME
+fi
+unset ACORN_USER_ZDOTDIR
+[ -f "$ZDOTDIR/.zshrc" ] && source "$ZDOTDIR/.zshrc"
+
+_acorn_osc7() {
+    printf '\e]7;file://%s%s\e\\' "${HOST:-${HOSTNAME-localhost}}" "$PWD"
+}
+# precmd_functions is the canonical "run before each prompt" hook list in
+# zsh. typeset -gaU keeps it global, array-typed, and dedup'd so a repeated
+# `source` of this file (e.g. `exec zsh`) doesn't stack duplicate emitters.
+typeset -gaU precmd_functions
+precmd_functions+=(_acorn_osc7)
+# Emit once immediately so the host gets the spawn cwd before the user runs
+# anything — without this, the icon waits for the first prompt.
+_acorn_osc7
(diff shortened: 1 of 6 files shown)


im-ian (commented):
## Self-review

**Decision**: APPROVE — no critical/high findings.

### Validation

- `bun run typecheck` pass
- `bun run test` pass (123 / 124, 1 skipped)
- `cargo clippy` clean for new code
- `cargo test shell_init` pass (2/2)
- Live-tested: OSC 7 emits on every prompt, backend classifies, `cd` into / out of a worktree updates the icon immediately (verified via webview console + UI on a real PTY)

### Findings

#### Medium

**No unit test for `is_path_linked_worktree`.** Trivial to add (temp dir + `Repository::init` + `worktree`), and it's a public Tauri command — would catch a regression if `Repository::discover` semantics changed. Low cost.

#### Low

1. **User's `.zshrc` could short-circuit our hook.** If their config calls `exit` or `return` before reaching the end of our rc, `precmd_functions+=(_acorn_osc7)` never runs and the OSC path silently degrades. Mitigation: the #156 focus-event refresh remains as fallback, so the icon still updates on app focus. Worth noting in CLAUDE.md when the file is touched.

2. **Daemon spawn path inherits the env.** `effective_env` is built before the `state.daemon_bridge.is_enabled()` branch in `pty_spawn`, so `ZDOTDIR` reaches the daemon-routed PTY too. Verified by inspection — not separately E2E tested.

3. **OSC 7 could be spoofed locally.** A child process could `printf '\e]7;file:///some/path\e\\'` and the icon would follow. Read-only consequence (no fs write), but the icon would lie about cwd. Not a security issue — local processes already share the PTY trust boundary.

4. **`.zlogin` / `.zprofile` skipped** when `ZDOTDIR` is overridden. PTY is interactive non-login on Acorn so `.zprofile` wouldn't have run anyway; `.zlogin` (login shells) likewise. No behavioral regression for the common case.

### Files reviewed

| File | Change |
|---|---|
| `src-tauri/shell-init/zshrc` | New — bundled rc with `precmd` OSC 7 hook |
| `src-tauri/src/shell_init.rs` | New — `ensure_shell_init_dir`, mirrors `agent_shim` |
| `src-tauri/src/commands.rs` | `is_path_linked_worktree` + `ZDOTDIR` env injection in `pty_spawn` |
| `src-tauri/src/lib.rs` | Register module + command |
| `src/lib/api.ts` | `isPathLinkedWorktree` wrapper |
| `src/components/Terminal.tsx` | xterm OSC 7 handler |

### Notes for the reviewer

Stacks on #156 — merging #156 first auto-retargets this PR to main. Bash and fish are intentionally out of scope: bash will use env `PROMPT_COMMAND` directly in a follow-up; fish emits OSC 7 natively.

PR #167, opened 2026-05-13
feat(claude-resume): focus-time modal in place of shim auto-resume

im-ian:
## Summary

PR #132 wired the claude shim to auto-inject `--session-id $ACORN_RESUME_TOKEN` on every invocation so a quit/reopen of Acorn would resume the conversation. That works for the "Acorn restarted" case, but the side effect is that every claude run in a long-lived worktree appends to the same JSONL forever — local transcripts are already at 3.8 MB / hundreds of turns, and each new run loads the whole history regardless of whether the user is starting a fresh topic. This PR replaces the always-on injection with a focus-time modal that surfaces the previous conversation explicitly and lets the user decide.

## Design

| Scenario | Before (PR #132) | This PR |
| --- | --- | --- |
| Fresh `claude` invocation | Auto-resumes the Acorn session's JSONL | Fresh conversation; the modal will offer to resume on next focus |
| Crash / unintended exit | Auto-resumes | Modal pops on next focus; user decides |
| Bloating context | Builds up indefinitely | Bounded — only intentional resumes accumulate |
| User control | Implicit | Explicit; visible UUID; clipboard escape hatch |
| Claude native `--resume` UX | Bypassed by shim | Used directly (the modal types `claude --resume <uuid>` for the user) |

The modal appears when:
1. `<state_dir>/claude.id` exists (the shim wrote it after a previous claude run), and
2. `claude.id != claude.id.acknowledged`, and
3. `claude` is not currently running in the session's PTY descendant tree (backend gate via `commands::claude_is_running_in_session`).

Three actions, all of which acknowledge so the same UUID does not re-pop:

- **이어하기** — `api.ptyWrite(sessionId, "claude --resume <uuid>\n")`. Claude's native `--resume` handles the rest.
- **ID 복사** — `navigator.clipboard.writeText(uuid)` + toast. Useful when the user wants to resume later or paste the id into another tool.
- **취소** — types two `#`-prefixed shell-comment lines into the PTY so the resume command stays visible in scrollback for later. Comments are inert if the user mashes Enter.

A small assistant-message preview (first ~90 chars of the last `assistant` text segment from the JSONL tail) is included so users can recognize *which* conversation they're being offered.

## Shim changes

`src-tauri/shims/claude.sh` no longer reads `ACORN_RESUME_TOKEN`. It snapshots `~/.claude/projects/<slug>/*.jsonl` before and after a bare-flag run, captures the new UUID, and writes it to `$ACORN_AGENT_STATE_DIR/claude.id`. Explicit `--resume` / `--continue` / `--session-id` from the user always passes through untouched. The env var stays in the PTY env so user scripts that addressed the session by UUID keep working.

## Notes

- This is also a step toward decoupling claude's transcript UUID from Acorn's session UUID — `AgentDetection.claude` and `transcript_watcher` are now the source of truth for "which JSONL belongs to this session," not a reused id.
- A separate bug is worth tracking: any user `.zshrc` that prepends `~/.local/bin` (or anywhere ahead of the shim dir) bypasses the shim entirely. That neutralises both the old auto-resume and this new capture path; surfacing that bypass (e.g. via the Settings → Sessions panel) is a follow-up.

## Test plan

- [x] `bun run typecheck` — clean
- [x] `bun run test` — 22 files, 186 passed / 1 skipped (includes new `ClaudeResumeModal.test.tsx`, 5 cases)
- [x] `bun run test:e2e` — 49 passed (includes new `claude-resume-modal.spec.ts`, 2 cases covering 이어하기 and 취소 PTY-write payloads)
- [x] `cargo check --bins` — clean
- [x] `cargo test --lib agent_shim` — 10 passed (4 new: `claude_shim_captures_uuid_on_bare_run`, `claude_shim_passthrough_on_explicit_resume`, `claude_shim_passthrough_on_explicit_session_id`, `claude_resume_candidate_respects_acknowledgment`, `extract_last_assistant_preview_picks_most_recent_text`)
- [ ] Manual: spawn a session, run `claude`, exit, open a different session, come back — modal pops; 이어하기 → claude reattaches to the same JSONL
- [ ] Manual: 취소 path — verify the two `#`-prefixed lines render in scrollback and that mashing Enter does not execute them
- [ ] Manual: ID 복사 — confirm clipboard contents + toast
- [ ] Manual: open a session where `claude` is currently running — modal must NOT pop (descendant-process gate)
- [ ] Manual: after 이어하기, leave + return to the same session — modal must NOT re-pop for the same UUID


im-ian:
Closing in favor of a Rust-side filesystem watcher.

After live testing on macOS, the shim-based capture approach is bypassed in practice for the vast majority of users. macOS `/etc/zprofile` runs `path_helper`, which rebuilds `PATH` from `/etc/paths` + `/etc/paths.d/*` and pushes Acorn-prepended entries (including the shim dir) to the back. Any user `.zshrc` that further prepends to `PATH` (e.g. `export PATH="$HOME/.local/bin:$PATH"`) buries the shim even deeper. Result: the real `claude`/`codex` binary resolves first, the shim never runs, `claude.id`/`codex.id` never get written, and the focus-time modal never appears.

The same root cause silently breaks the shim features that already existed on `main`:

- claude `--session-id $ACORN_RESUME_TOKEN` injection — never happens, so `session_status::detect` falls back to descendant-process scanning.
- codex auto-resume — bare `codex` always starts a fresh session.

Extending shell-init to cover bash/fish/etc. doesn't fix the structural problem either — shim depends on PATH ordering, which is fundamentally fragile.

Replacement: a Rust watcher on `~/.claude/projects/**/*.jsonl` and `$CODEX_HOME/sessions/**/rollout-*.jsonl`. On a new transcript event, walk the live `claude`/`codex` process tree to find which Acorn session owns the PTY, then write `claude.id` / `codex.id` directly. No PATH dependency, no shell assumption. Codex gets the same modal pattern for symmetry; the `codex resume <uuid>` shim trick goes away.

Follow-up PR: `feat/agent-resume-watcher`.

PR #225, opened 2026-05-18
fix(status): mark unpaired live codex as input-ready

im-ian:
## Summary

This PR tightens session status detection for live Codex processes that have not yet produced or paired a transcript.

1. **Codex liveness fallback**: Reuses the captured PTY root pid to inspect descendant processes during status polling.
2. **Unpaired transcript handling**: Maps a live, unpaired Codex descendant from a generic running shell hint to `needs_input`, while preserving transcript-based classification once a Codex transcript is available.
3. **Regression coverage**: Adds focused unit tests for unpaired Codex, paired Codex, and non-Codex fallback behavior.

## Expected impact

| Area | Impact |
| --- | --- |
| User-visible status | Codex sessions that are live but lack a paired transcript are less likely to remain incorrectly idle/running when they are ready for input. |
| Reliability | Status polling has a conservative fallback for a known transcript-pairing gap without changing paired transcript semantics. |
| Regression risk | Low; the override only applies when there is no transcript and the live descendant process is Codex. |

## Design notes

The transcript classifier remains authoritative whenever a transcript is resolved. The new fallback only refines shell hints for unpaired live Codex descendants, which covers the observed gap without changing Claude behavior or transcript-backed Codex turns.

## Follow-up (not in this PR)

Keep investigating why some restarted sessions do not immediately reattach their daemon stream/root pid. That path is separate from this fallback because status polling cannot infer descendants until a session is attached.

## Test plan

- [x] `cargo test status_hint_tests --manifest-path src-tauri/Cargo.toml`
  - Passed: 3 tests.
- [x] `cargo test -p acorn-session status --manifest-path src-tauri/Cargo.toml`
  - Passed: 25 tests.
- [x] Local Tauri dev smoke test on `http://127.0.0.1:1430`
  - Passed: Vite returned HTTP 200 and Tauri dev launched with the current branch.

## Notes

The pre-existing daemon reattach gap can still leave a session without a root pid immediately after restart. This PR does not create that condition; it narrows the misclassification once a live Codex descendant is available to the status detector.

### DIFF
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -19,6 +19,7 @@ use crate::todos::{self, TodoItem};
 use crate::worktree;
 use acorn_session::scrollback;
 use acorn_session::status as session_status;
+use acorn_session::status::AgentKind as StatusAgentKind;
 use acorn_session::{Project, Session, SessionKind, SessionStatus};
 
 use serde::Serialize;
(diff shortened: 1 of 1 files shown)


im-ian (commented):
Self-review complete.

Findings: none blocking.

Checked the diff for status-classification regressions. The fallback only fires when all three conditions are true: no resolved transcript, shell hint is `Running`, and a live descendant basename resolves to `codex`. Transcript-backed Codex and Claude paths keep the existing classifier behavior.

Verified locally:
- `cargo test status_hint_tests --manifest-path src-tauri/Cargo.toml`
- `cargo test -p acorn-session status --manifest-path src-tauri/Cargo.toml`
- Tauri dev smoke on `http://127.0.0.1:1430`

PR #228, opened 2026-05-18
perf(terminal): reduce output and subscription overhead

im-ian:
## Summary
This reduces terminal output hot-path overhead and narrows React store subscriptions that were causing unnecessary UI work.

1. **PTY output transport:** route terminal bytes through a raw-byte Tauri Channel, while keeping the legacy base64 event path as a fallback.
2. **Daemon session parity:** apply the same output router to daemon-managed sessions and keep the PTY crate transport-agnostic through a host-provided sink.
3. **Store subscription scope:** replace broad session subscriptions in Pane, Sidebar, and TopBar with narrower selectors and a memoized session-id map.
4. **Terminal environment resilience:** strip inherited host color-policy env from PTY children and guard staged zsh forwarders against self-source recursion, including trailing-slash and symlink cases.
5. **Regression coverage:** add Rust and frontend coverage for the output router, terminal env handling, shell-init generation, and store selectors.

## Expected impact
| Area | Impact |
| --- | --- |
| Terminal output | Avoids base64 decode and global event fan-out on the normal PTY output path. |
| Daemon sessions | Uses the same output delivery behavior as local PTY sessions, with fallback compatibility preserved. |
| React rendering | Sidebar, Pane, and TopBar avoid re-rendering from unrelated session store updates. |
| Terminal colors | New PTY sessions ignore inherited `NO_COLOR`/color-force policy unless explicitly provided. |
| Dev shell startup | Staged zsh forwarders no longer recurse when `ACORN_USER_ZDOTDIR` resolves back to Acorn shell-init. |

## Design notes
- `pty_output` owns Channel subscribers per session and falls back to the legacy base64 event path when a raw subscriber is unavailable.
- `Terminal.tsx` subscribes with a Tauri Channel, decodes raw output first, keeps the legacy listener as compatibility fallback, and disables stale callbacks on unsubscribe.
- The PTY crate receives a host-provided output sink instead of knowing about Tauri transport details.
- `pty_env` drops inherited color-policy variables from the app host process while preserving explicitly supplied session env values.
- Shell-init staging now compares normalized/canonical zsh paths before sourcing user dotfiles so dev-generated App Support files do not self-source.

## Follow-up (not in this PR)
- Adaptive status polling, virtualized code/diff rendering, right-panel event invalidation, and watcher batching are being handled as separate branches/worktrees.

## Test plan
- [x] `pnpm test` passed.
- [x] `pnpm exec playwright test tests/e2e/sidebar.spec.ts tests/e2e/panes.spec.ts tests/e2e/smoke.spec.ts` passed.
- [x] `pnpm run build` passed.
- [x] `cargo check` passed.
- [x] `cargo test pty_output` passed.
- [x] `cargo test push_tail` passed.
- [x] `cargo test pty_env` passed.
- [x] `cargo test shell_init` passed.
- [x] Manual zsh self-source reproductions with trailing-slash and symlink `ACORN_USER_ZDOTDIR` passed.

## Notes
- `pnpm run build` still reports the existing Vite dynamic-import and large-chunk warnings.
- Manual `pnpm run tauri dev` launched locally; ANSI RED/GREEN terminal color output rendered correctly after inherited `NO_COLOR` was stripped.
- The production app bundle was not replaced while addressing the dev-only zsh recursion issue.

### DIFF
--- a/src-tauri/crates/acorn-pty/src/lib.rs
+++ b/src-tauri/crates/acorn-pty/src/lib.rs
@@ -6,9 +6,9 @@
 //!   * a process child handle used to kill the process
 //!   * a stop flag used to signal the reader task to exit cleanly
 //!
-//! Output bytes are read on a blocking thread, base64-encoded, and emitted to
-//! the frontend via Tauri events:
-//!   * `pty:output:{session_id}` — payload `{ "data": "<base64>" }`
+//! Output bytes are read on a blocking thread and forwarded through the
+//! host-provided output sink. The host can send raw bytes through a Tauri
+//! Channel and fall back to legacy events when needed.
 //!   * `pty:exit:{session_id}` — payload `{ "code": Option<i32> }`
 
 use std::collections::VecDeque;
@@ -51,6 +51,7 @@ pub enum PtyError {
 }
 
 pub type PtyResult<T> = Result<T, PtyError>;
+pub type PtyOutputSink = Arc<dyn Fn(&str, &Uuid, &[u8]) + Send + Sync + 'static>;
 
 /// Coarse classification of a shell-mode session's liveness, derived purely
 /// from "does the PTY child have any descendant processes right now?".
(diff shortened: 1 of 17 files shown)


im-ian (commented):
Self-review complete.

Findings: no blocking issues found.

Checked the raw PTY Channel path, daemon fallback behavior, Terminal subscription cleanup, store selector scope, and zsh self-source guards. The single Channel subscriber per session matches the current TerminalHost architecture, which renders one Terminal instance per session and moves its DOM target between panes.

Verification run before merge:
- `git diff --check origin/main...HEAD` passed.
- `pnpm test -- --run src/store.test.ts src/components/Terminal.test.tsx` passed for the matched test file, 49 passed / 1 skipped.
- `cargo test pty_output` passed.
- `cargo test pty_env` passed.
- `cargo test shell_init` passed.

PR #230, opened 2026-05-18
perf(status): adapt session polling cadence

im-ian:
## Summary
This replaces the 1-second global session-status poll with an adaptive scheduler that focuses work on sessions likely to change.

1. **Adaptive status scheduler:** poll the active session fastest, keep running/needs-input sessions on a moderate cadence, and move stable sessions to a slower safety interval.
2. **Subset status polling:** allow `pollSessionStatuses` to probe specific session ids instead of always scanning every session.
3. **Overlap control:** serialize status probes and coalesce queued requests so focus/session changes do not create overlapping backend work.
4. **Visibility and focus behavior:** pause while the document is hidden and trigger immediate checks when focus returns, the active session changes, or new sessions appear.
5. **Regression coverage:** add focused tests for interval decisions, subset polling, coalescing, and new-session queueing.

## Expected impact
| Area | Impact |
| --- | --- |
| Session status polling | Avoids a full session scan every second during normal use. |
| Active session feedback | Keeps the foreground session responsive with shorter adaptive intervals. |
| Background sessions | Keeps running sessions fresh while reducing work for idle/completed sessions. |
| UI/background load | Reduces repeated backend probes and React store writes from status polling. |

## Design notes
- `sessionStatusPolling` centralizes interval and due-session decisions so the scheduler stays testable outside React.
- `pollSessionStatuses(ids?)` still supports full polling when no ids are passed, preserving existing call sites.
- In-flight polling is serialized; new requests queue and coalesce instead of racing.
- The scheduler performs one immediate startup probe, then switches to adaptive subset probes.

## Follow-up (not in this PR)
- Runtime validation in a real Tauri window should confirm focus and document visibility transitions behave as expected across macOS app activation states.

## Test plan
- [x] `pnpm exec tsc --noEmit` passed.
- [x] `pnpm exec vitest run src/store.test.ts src/lib/sessionStatusPolling.test.ts` passed, 58 tests passed and 1 skipped.

## Notes
- This branch was rebased onto the current `origin/main` after #228 merged.
- The initial app load can still probe existing sessions once immediately; subsequent checks use adaptive subset polling.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -61,6 +61,11 @@ import {
 import { applyBackgroundVars, clearBackgroundVars } from "./lib/background";
 import { applyTheme, useThemes } from "./lib/themes";
 import { extractTabFromEvent } from "./lib/settings-events";
+import {
+  nextSessionStatusPollDelayMs,
+  selectDueSessionStatusPollIds,
+  selectImmediateSessionStatusPollIds,
+} from "./lib/sessionStatusPolling";
 import { useAppStore } from "./store";
 import type { TranslationKey, Translator } from "./lib/i18n";
 import type { SessionStatus } from "./lib/types";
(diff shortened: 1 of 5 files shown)


im-ian (commented):
Self-review complete.

Findings: no blocking issues found.

Checked the adaptive scheduler lifecycle, subset polling path, overlapping poll serialization, immediate focus/session-change probes, and stale session cleanup. The added tests cover interval selection, subset polling, queued overlap behavior, full-poll coalescing, and sessions added during an active poll.

Verification before merge:
- `git diff --check origin/main...HEAD` passed.
- `pnpm exec tsc --noEmit` passed.
- `pnpm exec vitest run src/store.test.ts src/lib/sessionStatusPolling.test.ts` passed.

PR #233, opened 2026-05-18
perf(diff): virtualize large code and diff views

im-ian:
## Summary
This switches large code and diff views to windowed rendering while preserving the existing full-render path for small files.

1. **Shared virtual line renderer:** add a `VirtualizedLines` component backed by `@tanstack/react-virtual` for large line-oriented views.
2. **Code viewer virtualization:** render large files through the virtualized path while keeping line numbers, highlighted HTML, and horizontal sizing behavior.
3. **Diff view virtualization:** virtualize large unified/split diff line lists while preserving markers, line numbers, and highlight output.
4. **Small file compatibility:** keep the existing non-virtualized rendering path below the line threshold to avoid unnecessary complexity for normal files.
5. **Regression coverage:** add tests for virtualized code/diff rendering and preserve existing PR detail/diff test coverage.

## Expected impact
| Area | Impact |
| --- | --- |
| CodeViewer | Reduces DOM nodes and render work for large files. |
| DiffView | Reduces DOM pressure when opening large diffs. |
| DiffSplitView | Inherits virtualization without changing modal wiring. |
| Small files | Behavior remains on the existing full-render path. |

## Design notes
- Virtualization starts above 500 lines so small files keep their simpler current markup.
- Syntax highlighting still runs against loaded content before rendering; this PR targets DOM/render cost rather than highlighter CPU cost.
- The shared virtual renderer keeps line height stable and preserves horizontal overflow width for copy/selection behavior.

## Follow-up (not in this PR)
- Very large highlighted files may still spend CPU time before render; incremental or worker-based highlighting would be a separate optimization.

## Test plan
- [x] `pnpm exec tsc --noEmit` passed.
- [x] `pnpm exec vitest run src/components/CodeDiffVirtualization.test.tsx src/components/PullRequestDetailModal.test.ts src/components/PullRequestDetailModal.modal.test.tsx src/lib/diff.test.ts` passed, 20 tests.

## Notes
- This branch was rebased onto the current `origin/main` after #228 merged.
- `DiffViewerModal` was not changed directly; it receives the behavior through `DiffSplitView`.

PR #231, opened 2026-05-18
perf(right-panel): refresh from local invalidation events

im-ian:
## Summary
This reduces local right-panel polling by moving repo-local views to event invalidation plus a slower safety interval.

1. **Local invalidation scheduler:** add a small debounce helper for filesystem/git invalidation keys used by right-panel tabs.
2. **Commits and staged tabs:** replace short local polling loops with refreshes triggered by existing `acorn:fs-changed` metadata plus a 30-second safety interval.
3. **Todos refresh behavior:** keep transcript-driven todos responsive to active session/cwd changes and active PTY output, with the same slow safety interval as backup.
4. **Remote polling preserved:** leave GitHub PR, Actions, and workflow detail polling cadences unchanged because those are remote state checks.
5. **Regression coverage:** add tests for invalidation debouncing and key behavior.

## Expected impact
| Area | Impact |
| --- | --- |
| Right panel local tabs | Avoids repeated 2-3 second polling for local git and working-tree state. |
| Commits tab | Refreshes quickly when git metadata invalidates, with a slow fallback interval. |
| Staged tab | Refreshes from working-tree/git events instead of constant local polling. |
| Remote GitHub tabs | Behavior remains unchanged. |

## Design notes
- The implementation intentionally uses the existing `acorn:fs-changed` event surface so it does not overlap with the watcher batching backend work.
- The safety interval remains at 30 seconds to recover from missed filesystem events, linked worktree gitdir edge cases, or transcript-only updates.
- Todos are not treated as repo fs/git invalidations; they refresh on active context changes and debounced active terminal output.

## Follow-up (not in this PR)
- Once watcher batching/caps lands, right-panel invalidation can consume the richer batched event payload if needed.

## Test plan
- [x] `pnpm exec tsc --noEmit` passed.
- [x] `pnpm exec vitest run src/lib/right-panel-invalidation.test.ts` passed, 4 tests.

## Notes
- This branch was rebased onto the current `origin/main` after #228 merged.
- Linked worktree external gitdir changes may still depend on the 30-second safety interval until watcher event coverage is expanded.

PR #234, opened 2026-05-18
perf(files): batch watcher events with caps

im-ian:
## Summary
This batches filesystem watcher events before they reach the frontend so generated-file bursts do not fan out into per-file refresh work.

1. **Watcher event batching:** collect notify events in a 75ms backend batch window before emitting `acorn:fs-changed`.
2. **Event cap and overflow hint:** cap each payload at 256 paths and include a root/subtree refresh hint when the cap is exceeded or notify requests a rescan.
3. **Noisy path filtering:** ignore common generated and dependency directories such as `node_modules`, `target`, `dist`, and cache folders while preserving `.git` metadata events for git invalidation.
4. **Root containment guard:** drop watcher paths that normalize outside the active watched root.
5. **Frontend batch consumption:** update FileExplorer to consume the richer payload and refresh loaded directories, subtrees, or the root as appropriate.
6. **Regression coverage:** add Rust tests for noisy-path filtering, git metadata preservation, overflow subtree hints, and rescan root hints.

## Expected impact
| Area | Impact |
| --- | --- |
| File watcher bursts | Reduces frontend refresh storms during large generated-file changes. |
| FileExplorer | Refreshes loaded directories from batched paths and uses fallback subtree/root refresh on overflow. |
| Right panel invalidation | Continues receiving `.git` metadata paths for immediate commits/staged refresh. |
| Backend event volume | Emits bounded payloads with explicit overflow metadata. |
| Repo safety | Ignores common high-churn output folders and filters out-of-root paths. |

## Design notes
- The backend owns batching/cap behavior so all frontend consumers receive a bounded event stream.
- `FsChangePayload` remains backward-compatible for consumers that only read `paths`; new fields are optional on the TypeScript side.
- Overflow batches still include up to 256 representative paths, plus a refresh hint for broader recovery.
- FileExplorer no longer does its own event debounce for this path because the backend batch window now handles burst coalescing.
- `.git` is intentionally not filtered because the right panel uses git metadata events for local invalidation after short polling was removed.

## Follow-up (not in this PR)
- The ignore list includes common names like `build`, `dist`, and `target`; projects that intentionally keep source under those directory names may need a future configurable ignore policy.

## Test plan
- [x] `TAURI_SIDECAR_PROFILE=debug ./src-tauri/scripts/build-sidecar.sh` passed.
- [x] `cargo check --manifest-path src-tauri/Cargo.toml` passed.
- [x] `cargo test --manifest-path src-tauri/Cargo.toml fs_explorer` passed, 8 tests.
- [x] `pnpm exec tsc --noEmit` passed.
- [x] `git diff --check` passed.

## Notes
- This branch was rebased onto the current `origin/main` after #230, #231, and #233 merged.
- Local ignored build artifacts were produced during verification (`node_modules/`, `src-tauri/binaries/`, `src-tauri/gen/`, `src-tauri/target/`) and are not part of the commit.

### DIFF
--- a/src-tauri/src/fs_explorer.rs
+++ b/src-tauri/src/fs_explorer.rs
@@ -1,7 +1,9 @@
 use std::cmp::Ordering;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::path::{Component, Path, PathBuf};
+use std::sync::mpsc::{self, Receiver};
 use std::sync::Arc;
+use std::time::{Duration, Instant};
 
 use git2::{Repository, Status, StatusOptions};
 use ignore::gitignore::{Gitignore, GitignoreBuilder};
@@ -14,6 +16,25 @@ use crate::error::{AppError, AppResult};
 use crate::state::AppState;
 
 const EVENT_FS_CHANGED: &str = "acorn:fs-changed";
+const WATCH_BATCH_WINDOW: Duration = Duration::from_millis(75);
+const WATCH_EVENT_CAP: usize = 256;
+const WATCH_IGNORED_COMPONENTS: &[&str] = &[
+    "node_modules",
+    "target",
+    "dist",
+    "build",
+    "out",
+    ".next",
+    ".nuxt",
+    ".svelte-kit",
+    ".vite",
+    ".turbo",
+    ".cache",
+    ".parcel-cache",
+    ".pytest_cache",
+    "__pycache__",
+    "coverage",
+];
 
 #[derive(Debug, Clone, Serialize)]
 pub struct FileEntry {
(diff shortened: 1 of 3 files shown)


im-ian (commented):
Self-review complete.

Findings: fixed one issue before merge. The first version filtered `.git` events as noisy paths, which would have weakened the right-panel git invalidation added in #231 and pushed commits/staged refreshes back to the 30-second safety interval. I removed `.git` from the ignore list and updated the watcher test to assert git metadata events are preserved.

No remaining blocking issues found.

Checked backend batching/capping, overflow refresh hints, root containment, noisy path filtering, FileExplorer batch consumption, and compatibility with the existing right-panel `acorn:fs-changed` listener.

Verification before merge:
- `TAURI_SIDECAR_PROFILE=debug ./src-tauri/scripts/build-sidecar.sh` passed.
- `cargo check --manifest-path src-tauri/Cargo.toml` passed.
- `cargo test --manifest-path src-tauri/Cargo.toml fs_explorer` passed.
- `pnpm exec tsc --noEmit` passed.
- `git diff --check` passed.

PR #243, opened 2026-05-19
refactor(right-panel): centralize panel cache manager

im-ian:
## Summary
This PR moves right-panel cache ownership out of `RightPanel.tsx` and into a dedicated cache manager without changing the visible panel behavior.

1. **Centralized cache ownership:** Added `rightPanelCache` for Agent History, Commits, Staged, PR listings, workflow runs, in-flight de-dupe, and one-time project prefetch tracking.
2. **Project-scope pruning:** Added retained-repo tracking so caches for closed projects are removed while active sessions, worktrees, and workspace tabs remain retained.
3. **Stale in-flight guard:** Added repo-version checks so requests that resolve after their repo was pruned do not repopulate stale cache entries.
4. **Focused coverage:** Added cache-manager tests and reset the shared cache in the existing right-panel background loading tests.

## Expected impact

| Area | Impact |
| --- | --- |
| Right-panel behavior | No intended user-visible behavior change. Cached tabs should hydrate as before. |
| Reliability | Closed-project cache cleanup is explicit, and stale in-flight requests cannot re-add pruned entries. |
| Maintainability | Cache policy and in-flight de-dupe now live in one module instead of being spread across `RightPanel.tsx`. |
| Performance | Neutral to slightly improved through the same de-dupe behavior with clearer ownership. |

## Design notes

- The manager remains frontend process-memory only; it does not introduce persistence or a backend cache.
- `RightPanel` still decides which repos are open and when to prefetch. The manager only owns storage, de-dupe, pruning, and stale-result protection.
- Retained repos include open projects plus active session/worktree paths and workspace-tab repos so worktree-backed panels are not pruned just because their path differs from the project root.

## Follow-up (not in this PR)

- Consider exposing lightweight cache metrics for debugging prefetch behavior during project switching.
- Consider moving GitHub-origin cache ownership into the same manager later if we want one right-panel cache surface.

## Test plan

- [x] `pnpm test -- src/lib/right-panel-cache.test.ts src/components/RightPanel.test.tsx`
- [x] `pnpm typecheck`
- [x] `pnpm test`
- [x] `pnpm exec playwright test tests/e2e/right-panel.spec.ts --reporter=line`
- [x] `git diff --check`

## Notes

- The Playwright run still prints pre-existing E2E mock warnings for `load_status` and user theme loading; the right-panel tests passed despite that existing noise.

### DIFF
--- a/src/components/RightPanel.test.tsx
+++ b/src/components/RightPanel.test.tsx
@@ -49,6 +49,7 @@ vi.mock("./FileExplorer", () => ({
 }));
 
 import { api } from "../lib/api";
+import { rightPanelCache } from "../lib/right-panel-cache";
 import { useAppStore } from "../store";
 import { RightPanel } from "./RightPanel";
 
@@ -69,6 +70,7 @@ describe("RightPanel background tab loading", () => {
 
   beforeEach(() => {
     vi.useFakeTimers();
+    rightPanelCache.resetForTests();
     container = document.createElement("div");
     document.body.appendChild(container);
     root = createRoot(container);
@@ -109,6 +111,7 @@ describe("RightPanel background tab loading", () => {
     act(() => root.unmount());
     container.remove();
     vi.clearAllMocks();
+    rightPanelCache.resetForTests();
     vi.useRealTimers();
   });
 
--- a/src/components/RightPanel.tsx
+++ b/src/components/RightPanel.tsx
@@ -44,6 +44,7 @@ import { api, FS_CHANGED_EVENT, type FsChangePayload } from "../lib/api";
 import { cn } from "../lib/cn";
 import { openFileInEditor } from "../lib/editor";
 import { joinPath } from "../lib/paths";
+import { rightPanelCache } from "../lib/right-panel-cache";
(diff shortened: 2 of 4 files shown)


im-ian:
Self-review notes:

- Checked the cache ownership move against the existing right-panel behavior: `RightPanel` still controls tab mounting and prefetch timing, while `rightPanelCache` only owns storage, in-flight de-dupe, pruning, and stale-result protection.
- Looked specifically for hang risks: the manager does not add synchronous expensive work; pruning is bounded to current in-memory cache keys, and API work remains async through the existing frontend calls.
- Added repo-version guards so a request that resolves after project-close pruning returns to its caller but does not repopulate the pruned cache.
- Amended one small dependency cleanup after review: `retainRepos` now depends directly on the memoized retained repo list instead of a derived string key.

Verification after the amendment:

- `pnpm test -- src/lib/right-panel-cache.test.ts src/components/RightPanel.test.tsx`
- `pnpm typecheck`

Earlier full verification before the dependency cleanup:

- `pnpm test`
- `pnpm exec playwright test tests/e2e/right-panel.spec.ts --reporter=line`
- `git diff --check`

PR #247, opened 2026-05-19
fix(right-panel): hide git tabs outside repositories

im-ian:
## Summary

This keeps local git tabs tied to git repository availability while preserving the stricter GitHub-origin gate for GitHub-only tabs.

1. **Git probe split:** Added a cached git-repository probe alongside the existing GitHub-origin probe so local git UI can depend on `git init` without requiring an `origin`.
2. **Right panel visibility:** Hide `Staged` and `Commits` when the active project is not a git repo; keep them visible for git repos even when `origin` is missing or non-GitHub.
3. **Regression coverage:** Added unit and Playwright assertions for non-git projects and git repos without GitHub origin.

## Expected impact

| Area | Impact |
| --- | --- |
| Non-git projects | Code group shows `Files` only; `Staged` and `Commits` disappear with the GitHub group. |
| Git repos without GitHub origin | `Staged` and `Commits` stay available; GitHub group remains hidden. |
| Runtime git changes | Existing `.git` watcher invalidation now refreshes both git and GitHub-origin visibility asynchronously. |

## Design notes

- Reused the existing repo probe module instead of introducing a separate cache manager for this small visibility rule.
- The git-repo cache and GitHub-origin cache share the same generation invalidation, so `git init`, `.git` removal, and origin edits cannot leave one probe stale while the other refreshes.
- GitHub tab prefetch still requires the GitHub-origin probe to pass; local git tabs only require `is_git_repository`.

## Test plan

- [x] `pnpm test -- src/components/RightPanel.test.tsx`
- [x] `pnpm typecheck`
- [x] `pnpm exec playwright test tests/e2e/right-panel.spec.ts --reporter=line`
- [x] `pnpm test`
- [x] `git diff --check`

## Notes

- Playwright prints the existing `NO_COLOR` / `FORCE_COLOR` warning during the e2e run; it did not fail the suite.

### DIFF
--- a/src/components/RightPanel.test.tsx
+++ b/src/components/RightPanel.test.tsx
@@ -120,7 +120,7 @@ describe("RightPanel background tab loading", () => {
       activeProject: REPO,
       activeSessionId: null,
       activeTabId: null,
-      rightTab: "history",
+      rightTab: "commits",
       workspaceTabs: {},
       prAccountByRepo: {},
     });
@@ -203,7 +203,7 @@ describe("RightPanel background tab loading", () => {
     expect(mockApi.listWorkflowRuns).toHaveBeenCalledWith(REPO_B, 50);
   });
 
-  it("hides GitHub tabs for projects that are not git repositories", async () => {
+  it("hides git-backed tabs for projects that are not git repositories", async () => {
     mockApi.isGitRepository.mockResolvedValue(false);
 
     await act(async () => {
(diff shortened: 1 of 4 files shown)


im-ian (commented):
Self-review completed. I re-read the probe/cache split and RightPanel visibility path. The important edge cases are covered: non-git projects hide GitHub plus Staged/Commits, git repos without GitHub origin keep Staged/Commits, and dotgit_changed invalidates both git and GitHub-origin probes through the same generation counter. Local verification passed: pnpm test -- src/components/RightPanel.test.tsx, pnpm typecheck, pnpm exec playwright test tests/e2e/right-panel.spec.ts --reporter=line, pnpm test, and git diff --check.

PR #250, opened 2026-05-19
fix(paths): profile data dirs + shell-init env guard

im-ian:
## Summary

This moves dev/prod runtime isolation under a stable Acorn app data root and prevents Acorn shell-init dirs from being reused as user zsh config.

1. **Shared profile paths:** Added `acorn-paths` so app persistence, daemon paths, and `acorn-ipc` fallback sockets resolve through `ACORN_DATA_DIR`, then `ACORN_PROFILE`, then debug=`dev` / release=`prod`.
2. **Runtime env propagation:** The app now passes the resolved `ACORN_DATA_DIR` into spawned daemons and PTY children so release-built sidecars and shell commands stay on the same profile.
3. **Shell-init recursion guard:** PTY spawn now treats any Acorn-managed `shell-init` directory as internal and falls back to `$HOME` for `ACORN_USER_ZDOTDIR`, avoiding prod/dev staged `.zshenv` recursion.
4. **Prod file carry-forward:** Existing root-level prod `sessions.json` and `projects.json` are copied into `profiles/prod` on first access when the profile files do not already exist.
5. **Docs:** README and control-session docs now describe profile data dirs and the `ACORN_PROFILE` / `ACORN_DATA_DIR` override model.

## Expected impact

| Area | Before | After |
| --- | --- | --- |
| Dev/prod data isolation | Debug builds used a separate `io.im-ian.acorn-dev` app dir. | All builds use `io.im-ian.acorn/profiles/<profile>`, with debug defaulting to `dev` and release to `prod`. |
| IPC/daemon alignment | App, IPC fallback, and daemon each duplicated path logic. | Shared resolver keeps persistence, IPC socket, daemon sockets, pid/log, and shell-init in one profile. |
| zsh startup reliability | A sibling Acorn shell-init dir could be mistaken for user dotfiles and recurse. | Acorn shell-init dirs are detected and skipped as user `ZDOTDIR` targets. |
| Existing prod sessions/projects | Root-level prod files would otherwise be missed after the profile move. | Missing `profiles/prod` files are seeded from the legacy root files without overwriting newer profile data. |
| Developer docs | Runtime data layout was implicit in code comments. | Source-build and control-session docs mention profile dirs and env overrides. |

## Design notes

`ACORN_DATA_DIR` remains the strongest override for tests and explicit launches. `ACORN_PROFILE` is validated to simple path-safe profile names before it becomes a `profiles/<name>` segment, so values like `../prod` are rejected instead of traversing out of the data root.

The daemon spawn path passes `ACORN_DATA_DIR` rather than relying on the sidecar build profile. This matters because bundled sidecars may be release-built even when the host app is running a dev profile.

Legacy prod migration is intentionally limited to app-owned `sessions.json` and `projects.json`; sockets, pid files, logs, staged shell-init files, and daemon metadata can be recreated under the new profile directory. Explicit `ACORN_DATA_DIR` overrides skip the legacy copy path so isolated test/custom dirs stay isolated.

## Follow-up (not in this PR)

1. Consider a cleanup path for stale legacy root files once the profile layout has shipped and settled.

## Test plan

- [x] `cargo test -p acorn-daemon paths::tests::profile_env_selects_profile_subdir_under_acorn_app_dir` failed before implementation with `acorn-dev`, then passed after the resolver change.
- [x] `cargo test -p acorn persistence::tests::copies_legacy_persistence_file_only_when_profile_missing` failed before migration implementation, then passed after adding the carry-forward helper.
- [x] `pnpm run dev:sidecar` completed and staged local sidecar binaries for Tauri build-script checks.
- [x] `cargo test -p acorn-paths` passed: 3 tests.
- [x] `cargo test -p acorn-daemon paths::tests` passed: 3 tests.
- [x] `cargo test -p acorn-ipc socket_path::tests` passed: 2 tests.
- [x] `cargo test -p acorn persistence::tests` passed: 5 tests.
- [x] `cargo test -p acorn shell_init::tests::detects_acorn_shell_init_dirs` passed.
- [x] `env -u ZDOTDIR -u ACORN_USER_ZDOTDIR ACORN_PROFILE=dev pnpm run tauri dev` launched successfully and logged `profiles/dev` paths.

## Notes

`pnpm run dev:sidecar` created ignored files under `src-tauri/binaries/` so the Tauri build script could validate externalBin paths locally. Those generated binaries are not part of the PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,12 @@ pnpm run tauri build    # 프로덕션 빌드
 
 > ℹ️ `pnpm run tauri dev` / `tauri build`는 Tauri의 `externalBin` 규약에 따라 `src-tauri/binaries/acorn-ipc-<target-triple>`, `acornd-<target-triple>` 파일이 존재해야 시작합니다. 이 경로는 `.gitignore`에 포함돼 있어 fresh checkout(특히 `git worktree add`로 만들어진 worktree)에서는 비어 있고, 미리 빌드해두지 않으면 `resource path 'binaries/...' doesn't exist` 에러로 빌드가 실패합니다. `pnpm run build:sidecar`가 호스트 타깃에 맞는 두 바이너리를 빌드하고 올바른 위치에 stage합니다.
 
+개발 모드의 런타임 데이터는 설치된 앱과 분리되어
+`~/Library/Application Support/io.im-ian.acorn/profiles/dev/` 아래에 저장됩니다.
+프로덕션 빌드는 `profiles/prod/`를 사용합니다. 임시 격리가 필요하면
+`ACORN_PROFILE=<name>` 또는 `ACORN_DATA_DIR=/tmp/acorn-test`로 앱, IPC CLI,
+daemon이 볼 데이터 디렉터리를 바꿀 수 있습니다.
+
 ### 테스트
 
 ```bash
--- a/docs/CONTROL_SESSIONS.md
+++ b/docs/CONTROL_SESSIONS.md
@@ -49,6 +49,9 @@ layers that fire automatically every time a control-session PTY spawns:
    control session's PTY before any user code runs:
    - `ACORN_SESSION_ID` — this session's UUID, so `acorn-ipc` knows who
      is asking.
+   - `ACORN_DATA_DIR` — the resolved Acorn profile data directory. This
+     keeps bundled release sidecars aligned with the app's selected
+     profile.
    - `ACORN_IPC_SOCKET` — the canonical IPC socket path.
    - `ACORN_DAEMON_SOCKET` — the background daemon control socket path.
    - `PATH` — the directory containing the bundled `acorn-ipc` binary is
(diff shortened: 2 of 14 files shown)


im-ian (commented):
Self-review notes:

- Checked the profile resolver precedence: `ACORN_DATA_DIR` wins, `ACORN_PROFILE` is path-segment validated, and debug/release defaults remain deterministic.
- Caught and fixed a migration edge case during review: explicit `ACORN_DATA_DIR` overrides now skip legacy prod file copy so isolated test/custom dirs do not import real prod `sessions.json` or `projects.json`.
- Checked sidecar alignment: app-spawned daemon and PTY children receive the resolved `ACORN_DATA_DIR`, so release-built sidecars do not fall back to the wrong profile.
- Checked zsh recursion path: `ACORN_USER_ZDOTDIR` now falls back to `$HOME` when the incoming value points at any Acorn-managed `shell-init` dir, not just the current staged dir.

No blocking findings remaining from self-review.

im-ian (commented):
Updated self-review after the migration follow-up:

Findings: none blocking.

Checked again after the latest commits:

- `acorn-paths` keeps `ACORN_DATA_DIR` as the strongest override and validates `ACORN_PROFILE` before joining it under `profiles/<name>`.
- App persistence now carries legacy root-level prod `sessions.json` and `projects.json` into `profiles/prod` without overwriting existing profile files.
- Explicit `ACORN_DATA_DIR` overrides skip legacy prod migration, so tests/custom isolated dirs do not accidentally import real prod state.
- Daemon spawn and PTY child env both receive the resolved `ACORN_DATA_DIR`, which keeps release-built sidecars aligned with the app-selected profile.
- Shell-init recursion guard now detects any Acorn-managed `shell-init` directory by staged file contents, not only the current profile path.

Residual non-blocking note: existing `io.im-ian.acorn-dev` developer state is not migrated. That is acceptable for this PR as dev-local continuity, and the PR body tracks cleanup/migration follow-up separately.

Verification reviewed: GitHub CI Frontend and E2E are green, and the targeted Rust tests listed in the PR body pass locally.

PR #264, opened 2026-05-20
fix(worktree): prompt before deleting linked session worktrees

im-ian:
## Summary
Close-session cleanup now uses the same recorded-worktree signal as the UI worktree glyph, while keeping transient live-cwd changes out of destructive decisions.

1. **Prompt linked worktree sessions** — treat `session.isolated || session.in_worktree` as the recorded worktree condition for tab-close confirmation, so Acorn-managed isolated sessions and adopted/linked worktree sessions both offer keep/delete choices.
2. **Keep transient cwd non-destructive** — leave `liveInWorktree` as a display-only signal, so a user or agent temporarily `cd`ing into another worktree does not make that path a removal target.
3. **Delete by actual path** — replace session-name-derived worktree deletion with path-based git worktree lookup/removal, covering suffixed worktree names and linked-worktree project roots.
4. **Clarify copy and tests** — label non-isolated linked sessions as linked git worktrees, add focused React/unit tests, and cover backend deletion by actual path.

## Expected impact
| Scenario | Before | After |
| --- | --- | --- |
| Acorn isolated session close | Prompted to keep/delete worktree | Unchanged |
| `claude -w` / adopted linked worktree close | Could be treated like a plain session | Prompts to keep/delete recorded worktree |
| Agent temporarily `cd`s into a worktree | Worktree icon may show while live | Still display-only; no delete prompt solely from live cwd |
| Worktree name differs from session name | Delete could miss the actual worktree | Delete resolves the registered git worktree by path |

## Design notes
- `hasRecordedWorktree(session)` centralizes the static icon/removal signal and intentionally excludes the live PTY cwd map.
- Non-isolated linked worktrees default Enter to keeping the worktree, since those paths may be user-managed even when they are valid linked worktrees.
- Backend deletion validates the target as a linked worktree and matches against registered git worktree paths before removing the directory and pruning metadata.
- Project close now counts recorded linked worktrees, but avoids making linked non-isolated worktree deletion the keyboard-default action.

## Follow-up (not in this PR)
- Surface a more explicit warning for project-root linked worktrees if user testing shows the current linked-worktree wording is too subtle.

## Test plan
- [x] `pnpm run build:sidecar` — staged `acorn-ipc-aarch64-apple-darwin` and `acornd-aarch64-apple-darwin` in the clean PR worktree
- [x] `pnpm run typecheck` — passes in the clean PR worktree
- [x] `pnpm vitest run src/components/RemoveSessionDialog.test.tsx src/lib/sessionWorktree.test.ts src/store.test.ts src/lib/worktreeAdoption.test.ts` — 70 passed, 1 skipped
- [x] `cargo test -p acorn --lib commands::tests` — 16 passed
- [x] `cargo test -p acorn --lib worktree::tests` — 3 passed
- [x] `git diff --check` — passes

## Notes
- The implementation was prepared in a clean `/tmp/acorn-worktree-close-prompts` worktree from `origin/main` so pre-existing local dirty files on `feat/agent-hook-runtime-wrappers` are not included.

### DIFF
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -862,9 +862,8 @@ pub async fn remove_project(
             .collect();
         for session in session_ids {
             state.pty.kill(&session.id).ok();
-            if drop_worktrees && session.isolated {
-                let safe_name = sanitize_worktree_name(&session.name);
-                worktree::remove_worktree(&session.repo_path, &safe_name).ok();
+            if drop_worktrees {
+                remove_linked_worktree_at_path(&session.repo_path, &session.worktree_path).ok();
             }
             state.sessions.remove(&session.id).ok();
         }
@@ -886,9 +885,8 @@ pub async fn remove_session(
     if let Ok(dir) = persistence::data_dir() {
         scrollback::delete(&dir, &id.to_string()).ok();
     }
-    if session.isolated && remove_worktree.unwrap_or(false) {
-        let safe_name = sanitize_worktree_name(&session.name);
-        worktree::remove_worktree(&session.repo_path, &safe_name).ok();
+    if remove_worktree.unwrap_or(false) {
+        remove_linked_worktree_at_path(&session.repo_path, &session.worktree_path).ok();
     }
     state.sessions.remove(&id)?;
     persist(&state);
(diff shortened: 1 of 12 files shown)


im-ian (commented):
Self-review complete. I rechecked the destructive path: the frontend prompt now uses only recorded worktree state (), while live PTY cwd remains display-only. Backend deletion resolves the actual registered git worktree path before removing/pruning, with coverage for session-name mismatch and project-root linked worktrees. No blocking findings from the self-review.

im-ian:
Self-review note correction: recorded worktree state means session.isolated OR session.in_worktree. The earlier review comment lost that expression because the shell interpreted backticks.

PR #267, opened 2026-05-20
fix(terminal): transcript-driven remote-control redraw

im-ian:
## Summary
Remote-control terminal redraw now keys off transcript JSONL advancement instead of racing hook status transitions.

1. **Backend transcript watcher:** add a 300ms polling watcher that tracks live agent transcript path, size, and mtime per session, refreshing transcript mappings every 2s and emitting `acorn:agent-transcript-advanced` when a transcript advances.
2. **Frontend redraw wiring:** expose the new event payload type, refresh sessions from `App`, and repaint active terminals only when the advanced transcript belongs to that terminal session.
3. **PTY redraw nudge:** on transcript advancement, force a same-size PTY resize before repainting so Codex TUI can rewrite its current screen when remote-control output advanced the transcript without fresh PTY bytes.
4. **Hook helper hardening:** run Codex/Claude notify helpers through `/bin/sh` instead of PATH-dependent `bash`, and recognize Codex `SessionStart` hook payloads.
5. **Repaint policy tests:** add focused coverage for status-transition fallback redraws and transcript-advance redraw filtering.

## Expected impact
| Area | Impact |
| --- | --- |
| Remote-control reliability | Reduces blank/stale terminal redraws when status hooks arrive before assistant transcript writes finish, and handles transcript-only remote-control turns by nudging the TUI to redraw. |
| Terminal scrollback | No synthetic transcript text is injected; Acorn only refreshes existing xterm state and asks the PTY app to redraw via resize. |
| Backend overhead | Transcript path remapping is limited to every 2s; normal polling stats known transcript files every 300ms. |

## Design notes
- The watcher treats the first non-empty transcript cursor as an advancement so already-started remote-control turns can trigger a redraw once the transcript is discoverable.
- Missing live transcripts or missing files clear the cursor for that session, avoiding stale path tracking across agent rotations.
- Status-based repaint remains as a fallback for `running -> needs_input|failed|completed`, but transcript advancement is the reliable redraw signal.
- The PTY resize is intentionally same-size and text-free: it does not inject transcript content into scrollback, but it gives TUIs such as Codex a standard redraw signal.

## Follow-up (not in this PR)
- Transcript preview snapshot APIs remain separate from redraw. This PR intentionally does not parse or surface transcript text.

## Test plan
- [x] `pnpm test src/lib/terminalAttentionRepaint.test.ts` - passed, 6 tests.
- [x] `pnpm typecheck` - passed.
- [x] `cd src-tauri && cargo test -p acorn --lib agent_transcript_watcher -- --nocapture` - passed, 2 tests.
- [x] `cd src-tauri && cargo test -p acorn --lib agent_wrappers -- --nocapture` - passed, 6 tests.
- [ ] `cd src-tauri && cargo fmt --check` - failed on pre-existing rustfmt diffs in files not touched by this PR, including `crates/acorn-session/src/status.rs`, `src/commands.rs`, `src/fs_explorer.rs`, `src/persistence.rs`, and `src/pty_env.rs`.
- [x] `rustfmt --edition 2021 --check src-tauri/src/agent_transcript_watcher.rs src-tauri/src/agent_wrappers.rs` - passed.
- [x] `git diff --check` - passed.

## Notes
- The `cargo fmt --check` failure is not caused by this PR; the reported diffs are outside the changed file set.

### DIFF
--- a/src-tauri/crates/acorn-daemon/src/server.rs
+++ b/src-tauri/crates/acorn-daemon/src/server.rs
@@ -290,13 +290,17 @@ impl Daemon {
                 target_session_id,
                 data_b64,
             } => match base64_decode(&data_b64) {
-                Ok(bytes) => match self.pty.write(&target_session_id, &bytes) {
-                    Ok(()) => ControlResult::Ack,
-                    Err(err) => ControlResult::Error {
-                        code: io_error_to_code(&err),
-                        message: err.to_string(),
-                    },
-                },
+                Ok(bytes) => {
+                    let scrollback = self.pty.scrollback_snapshot(&target_session_id);
+                    let bytes = normalize_input_for_terminal_mode(bytes, scrollback.as_deref());
+                    match self.pty.write(&target_session_id, &bytes) {
+                        Ok(()) => ControlResult::Ack,
+                        Err(err) => ControlResult::Error {
+                            code: io_error_to_code(&err),
+                            message: err.to_string(),
+                        },
+                    }
+                }
                 Err(msg) => ControlResult::Error {
                     code: ErrorCode::Invalid,
                     message: msg,
@@ -476,6 +480,8 @@ impl Daemon {
                     match frame {
                         StreamFrame::Input { data_b64 } => {
                             if let Ok(b) = base64_decode(&data_b64) {
+                                let scrollback = pty_for_input.scrollback_snapshot(&session_id);
+                                let b = normalize_input_for_terminal_mode(b, scrollback.as_deref());
                                 let _ = pty_for_input.write(&session_id, &b);
                             }
                         }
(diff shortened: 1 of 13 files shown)


im-ian:
Related follow-up: #268

This PR covers the redraw-trigger side of the problem by watching transcript advancement and nudging the PTY/TUI to repaint without injecting transcript text into xterm scrollback.

#268 tracks the remaining product/architecture gap: remote turns can advance the provider transcript while producing no local PTY bytes at all. In that case, xterm repaint/resize/serialize cannot show the missing content because it was never written to the terminal buffer. The proposed direction there is to keep xterm PTY-only and surface transcript-only activity through a separate transcript-aware side channel, such as a badge, adjacent drawer, or session timeline, instead of synthesizing terminal output.

PR #295, opened 2026-05-21
feat(sessions): add worktree auto-delete removal setting

im-ian:
## Summary

Adds a session setting for users who want worktree-backed closes to always delete the worktree without another prompt.

1. **Settings model:** adds `sessions.autoDeleteWorktrees`, defaulting off so existing keep/delete confirmation behavior is preserved.
2. **Removal flow:** skips the remove-session dialog for linked or isolated worktree sessions when the setting is enabled and calls `remove_session` with `removeWorktree: true`.
3. **Settings UI + locales:** exposes the toggle in Settings → Sessions with English and Korean copy, and updates related hints.
4. **Regression coverage:** adds Vitest coverage for persisted settings/UI patching and Playwright coverage for the prompt-skipping removal path.

## Expected impact

| Area | Impact |
| --- | --- |
| User-visible behavior | Users can opt into always deleting worktrees on session close, avoiding the keep/delete prompt. |
| Safety | Default remains off, so existing users keep the confirmation behavior until they choose otherwise. |
| Reliability | Auto-close-on-exit routes through the same removal policy as manual tab close. |

## Design notes

The setting lives alongside the existing session removal preferences because the choice is frontend policy over the existing backend `removeWorktree` flag. `App.tsx` also passes `null` into `RemoveSessionDialog` for auto-handled removals so the dialog cannot flash for a frame before the effect removes the session.

## Test plan

- [x] `pnpm exec vitest run src/lib/settings.test.ts src/components/SettingsModal.test.tsx` — 2 files / 36 tests passed.
- [x] `pnpm exec playwright test tests/e2e/session-lifecycle.spec.ts` — 3 tests passed.

## Notes

Playwright still prints existing mocked-dev warnings for `load_status` and user-theme loading during this spec; the tests passed and those warnings are not introduced by this PR.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -147,12 +147,24 @@ function App() {
   );
   const removeSession = useAppStore((s) => s.removeSession);
   const removeProject = useAppStore((s) => s.removeProject);
+  const confirmRemoveSession = useSettings(
+    (s) => s.settings.sessions.confirmRemove,
+  );
+  const autoDeleteWorktrees = useSettings(
+    (s) => s.settings.sessions.autoDeleteWorktrees,
+  );
   const pendingRemove = sessions.find((s) => s.id === pendingRemoveId) ?? null;
   const pendingProject =
     projects.find((p) => p.repo_path === pendingRemoveProject) ?? null;
   const pendingProjectSessions = pendingProject
     ? sessions.filter((s) => s.repo_path === pendingProject.repo_path)
     : [];
+  const pendingRemoveRecordedWorktree =
+    pendingRemove !== null && hasRecordedWorktree(pendingRemove);
+  const pendingRemoveSkipsDialog =
+    pendingRemove !== null &&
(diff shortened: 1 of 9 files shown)


im-ian (commented):
Self-review complete.

Findings: none blocking.

Checked:
- `sessions.autoDeleteWorktrees` defaults to `false`, so existing users keep the prompt until they opt in.
- The auto-delete path is limited to `hasRecordedWorktree(session)` and passes `removeWorktree: true` to the existing backend removal API.
- `RemoveSessionDialog` receives `null` for auto-handled removals, preventing a transient keep/delete dialog render.
- Manual close and close-on-exit both route through the same session-removal policy.
- Coverage includes persisted settings, Settings UI patching, and the Playwright prompt-skipping removal path.

Residual risk: the new option is intentionally destructive once enabled; the UI copy calls out that the worktree is deleted from disk.

PR #299, opened 2026-05-21
fix(sessions): remove worktree sessions optimistically

im-ian:
## Summary

Removing a session with worktree deletion now updates the UI immediately while the backend cleanup continues.

1. **Optimistic session removal:** Drop the session from the frontend store before `remove_session` finishes so the terminal/tab disappears as soon as the user confirms deletion.
2. **Pane and cache cleanup:** Reconcile panes immediately and clear per-session live worktree / pending terminal input cache entries.
3. **Failure recovery:** If backend deletion fails, refresh from the backend to restore the real session list and preserve the error message.
4. **Regression coverage:** Add store tests for pending backend deletion and optimistic-removal rollback.

## Expected impact

| Area | Expected impact |
| --- | --- |
| User-visible behavior | The remove-worktree confirmation no longer leaves the old terminal visible during slow filesystem cleanup. |
| Reliability | Failed backend deletion refreshes from source of truth so the UI can recover the session. |
| Performance | No backend performance change; perceived latency is reduced by updating local state first. |

## Design notes

The backend command still owns PTY shutdown, scrollback deletion, worktree removal, session persistence, and local project mirror cleanup. The frontend only removes the tab optimistically, then performs the same `refreshAll()` after the backend returns to reconcile with the persisted state.

On failure, the store refreshes before setting the error so a successful recovery refresh does not clear the deletion failure message.

## Test plan

- [x] `pnpm test -- src/store.test.ts src/components/RemoveSessionDialog.test.tsx` — 2 files passed, 68 tests passed, 1 skipped.

### DIFF
--- a/src/store.test.ts
+++ b/src/store.test.ts
@@ -121,6 +121,20 @@ async function seed(
   await useAppStore.getState().refreshAll();
 }
 
+function deferred<T>(): {
+  promise: Promise<T>;
+  resolve: (value: T | PromiseLike<T>) => void;
+  reject: (reason?: unknown) => void;
+} {
+  let resolve!: (value: T | PromiseLike<T>) => void;
+  let reject!: (reason?: unknown) => void;
+  const promise = new Promise<T>((res, rej) => {
+    resolve = res;
+    reject = rej;
+  });
+  return { promise, resolve, reject };
+}
+
 beforeEach(() => {
   // resetAllMocks clears both call history *and* the queued
   // mockResolvedValueOnce return values; clearAllMocks leaves the queue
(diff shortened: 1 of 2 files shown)


im-ian (commented):
Self-review completed.

Findings: none.

Checked the optimistic removal path against the existing backend ownership boundary: PTY kill, scrollback cleanup, worktree deletion, session persistence, and local project mirror cleanup still stay in `remove_session`. The frontend only removes local UI state early, then reconciles via `refreshAll()`.

Also checked rollback behavior: backend deletion failure refreshes from source of truth before restoring the error message, so a failed worktree cleanup can bring the session back instead of leaving the UI permanently ahead of persisted state.

PR #339, opened 2026-05-27
feat(agents): filter history by provider

im-ian:
## Summary
Adds an agent provider filter to Agents History so mixed Claude/Codex session lists can be narrowed in place.

1. **History filter UI** — replace the History header count with the shared `Select` pattern used by GitHub Actions, with All agents, Codex, and Claude options.
2. **Provider-scoped rendering** — derive visible History rows from the selected provider while preserving existing refresh, empty, context-menu, resume, and trash flows.
3. **Localization and coverage** — add English/Korean labels plus focused Vitest and Playwright coverage for provider filtering.

## Expected impact
| Area | Impact |
| --- | --- |
| Agents History usability | Users can quickly isolate Codex or Claude sessions without changing the underlying history scan. |
| Existing resume/trash actions | No contract change; actions continue to operate on the rendered history item. |
| Backend/API behavior | No change; filtering is client-side over existing `AgentHistoryItem.provider` data. |

## Design notes
- The filter intentionally mirrors GitHub Actions' compact header `Select` so the control placement and sizing stay consistent across right-panel tabs.
- Filtering remains local UI state in `AgentHistoryTab`; cache contents and refresh behavior stay provider-agnostic.
- The provider type reuses the existing `AgentHistoryProvider` union, avoiding a parallel string contract.

## Test plan
- [x] `pnpm vitest run src/components/RightPanel.test.tsx` — 14 tests passed
- [x] `pnpm exec playwright test tests/e2e/right-panel.spec.ts` — 16 tests passed

### DIFF
--- a/src/components/RightPanel.test.tsx
+++ b/src/components/RightPanel.test.tsx
@@ -95,6 +95,14 @@ function buttonContaining(container: HTMLElement, text: string): HTMLButtonEleme
   return button;
 }
 
+function selectWithAria(container: HTMLElement, label: string): HTMLSelectElement {
+  const select = container.querySelector(`select[aria-label="${label}"]`);
+  if (!(select instanceof HTMLSelectElement)) {
+    throw new Error(`select labelled "${label}" not found`);
+  }
+  return select;
+}
+
 const labeledPullRequest = {
   number: 247,
   title: "Hide git tabs outside repositories",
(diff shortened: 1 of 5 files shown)


im-ian:

- CI is green: `Frontend` and `E2E` both passed.
- `feat` label is applied.

Holding merge intentionally because this is a feature PR and we want the pending SemVer patch/fix release to ship first without feature changes mixed into that patch line.

PR #360, opened 2026-05-29
fix(terminal): restore Cmd+V image paste fallback

im-ian:
## Summary
Terminal image paste now reads the macOS clipboard directly and uses one delayed compatibility fallback when native paste does not update the terminal.

1. **macOS clipboard snapshot** — adds a Tauri command that reads NSPasteboard text plus PNG/TIFF image payloads and exposes the data through the existing frontend API wrapper.
2. **App-owned Cmd+V routing** — replaces the macOS Edit > Paste item with an `acorn:paste` event, then routes it to the focused terminal, active session, or focused text input.
3. **Duplicate-resistant image fallback** — image paste waits briefly before Acorn-generated fallback work; terminal input/output during that window cancels the fallback.
4. **Agent-aware attachment behavior** — Claude and Codex receive the Ctrl+V PTY control path, while non-agent terminals save the image under app-local `clipboard-attachments` and insert an escaped file mention.
5. **Platform scope** — the custom paste menu is macOS-only, so non-macOS builds keep Tauri standard paste behavior.

## Expected impact
| Area | Before | After |
| --- | --- | --- |
| Image-only Cmd+V in Acorn | WKWebView could expose image clipboard types without a readable file, leaving Claude/Codex image paste stuck | Acorn reads NSPasteboard directly, then runs the provider-specific fallback if native paste does not update the terminal |
| Future native image paste support | A fallback could race with a later native terminal update | Fallback is delayed and cancelled on observed terminal input/output, so native success should remain single-registration |
| Claude/Codex sessions | Codex stayed on Ctrl+V fallback, Claude could still miss the image path | Both agents share the Ctrl+V/NSPasteboard-compatible fallback path |
| Other terminals | No Acorn-managed persisted image fallback | Image bytes are stored in app-local data and inserted as an escaped `@path` mention |
| Non-macOS paste | Risk from globally overriding `CmdOrCtrl+V` | Standard Tauri paste is preserved outside macOS |

## Design notes
- Text paste still uses xterm `term.paste()` so bracketed paste behavior remains unchanged.
- The fallback timer records terminal activity before native clipboard inspection. If a future native paste path produces input or output while Acorn reads the clipboard, the compatibility fallback is skipped.
- The backend command returns `supported: false` on non-macOS; the frontend custom paste menu only emits on macOS.
- Persisted fallback images live under `$APPLOCALDATA/clipboard-attachments`, keeping pasted screenshots out of project worktrees.

## Follow-up (not in this PR)
- Add a WKWebView image clipboard E2E once the test harness can seed an OS image clipboard reliably.
- Revisit Antigravity-specific UX if we want to prefer its native Ctrl+V media attachment path over the visible `@path` fallback.

## Test plan
- [x] `rustfmt --edition 2021 --check --config skip_children=true src-tauri/src/lib.rs src-tauri/src/clipboard.rs` passes
- [x] `cargo check` passes
- [x] `pnpm test src/lib/terminalPaste.test.ts src/lib/clipboardImageAttachment.test.ts` — 13 tests pass
- [x] GitHub CI Frontend and E2E pass on latest push

## Notes
- Full `cargo fmt -- --check` currently reports formatting diffs in pre-existing Rust files outside this PR scope; the Rust files changed here pass targeted rustfmt with child module traversal disabled.


im-ian:

- Required checks are green: `Frontend` passed.
- `fix` label is applied.

Holding merge intentionally because this ready-for-merge workflow stops before merging.

PR #378, opened 2026-06-03
feat(chat): empty composer prompts + pending timer polish

im-ian:
## Summary
This PR polishes the native chat empty state and running-response metadata so first-run chat mode feels more intentional and consistent.

1. **Empty composer prompt** — adds a stable per-render random title above the initial centered chat composer, using 10 Korean prompt options.
2. **Composer layout and motion** — expands the initial composer to 90% of the pane, removes the upward slide-in transition, and uses a simple fade-in/fade-out treatment.
3. **Pending response metadata** — moves the elapsed running time out of the `Running <provider>` bubble and into the same below-message metadata row used by completed messages.

## Expected impact
| Area | Before | After |
| --- | --- | --- |
| Empty chat state | Composer appeared narrower and slid upward on first load | Composer fills roughly 90% of the pane and fades in with a contextual title |
| Chat startup prompt | No headline text above the composer | One of 10 Korean starter prompts appears above the composer |
| Pending response duration | Timer appeared inline inside the pending assistant bubble | Timer appears below the bubble like normal chat metadata |

## Design notes
- The random empty-state title is selected once per `ChatPane` mount with lazy `useState`, so rerenders do not rotate text while the user is typing.
- The title fades out as soon as a message starts sending or existing messages are present.
- Pending messages now render a metadata row only for the running duration; copy/regenerate/edit/delete/fork actions remain hidden while a response is pending.

## Test plan
- [x] `pnpm typecheck` passes
- [x] `pnpm test src/components/ChatPane.test.tsx` passes — 29 tests
- [x] `pnpm run build` passes

## Notes
- `pnpm run build` still emits existing Vite chunk/dynamic-import warnings; no new build failure was introduced by this PR.

### DIFF
--- a/src/components/ChatPane.test.tsx
+++ b/src/components/ChatPane.test.tsx
@@ -905,6 +905,9 @@ describe("ChatPane", () => {
       "[data-chat-running-duration]",
     );
     expect(runningDuration?.textContent).toBe("2m 5s");
+    expect(
+      runningDuration?.parentElement?.hasAttribute("data-chat-message-meta"),
+    ).toBe(true);
   });
 
   it("updates elapsed response time while an assistant message is pending", async () => {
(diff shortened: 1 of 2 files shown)


im-ian:
Self-review note: reviewed the ChatPane diff and the pending-state rendering path. No blocking findings. Focused checks passed locally: `pnpm typecheck`, `pnpm test src/components/ChatPane.test.tsx`, and `pnpm run build`.

PR #384, opened 2026-06-05
fix(terminal): keep process probes off the main thread

im-ian:
## Summary
This keeps live terminal process probes from tying up the Tauri IPC path when Acorn resumes focus with active agent sessions.

1. **Offload live process probes.** Converts `detect_session_agent`, `pty_cwd`, and `pty_in_worktree_all` to async Tauri commands and runs their process-table work through `run_blocking`.
2. **Reduce FileExplorer probe fanout.** Stops running `detect_session_agent` on every app-store mutation while still refreshing the agent state when the file context menu opens.
3. **Preserve caller behavior.** Keeps frontend API contracts unchanged for the affected callers; the batched worktree probe still resolves to the same map shape after Tauri unwraps the command result.

## Expected impact
| Area | Before | After |
| --- | --- | --- |
| UI responsiveness | Focus/menu/output paths could synchronously scan the process table on the Tauri IPC path. | Process scans run on the blocking pool, reducing resume/focus stalls after remote or locked-screen work. |
| FileExplorer agent probes | Active-session agent detection ran on unrelated store updates. | FileExplorer tracks active-session changes and probes live agents on context-menu open. |
| Agent action freshness | On-demand detection stayed fresh but could block the app path. | Detection remains on-demand for menus while the blocking work is off the app path. |

## Design notes
- The agent detection path intentionally remains an on-demand live scan instead of using the watcher cache, because Fork/attach menu actions need a current process snapshot.
- `pty_repo_root` already used blocking work; this PR keeps that shape and only updates the shared `session_root_pid` helper to accept `AppState` directly.
- `pty_in_worktree_all` now returns `AppResult<HashMap<_, _>>` so it can use the same `run_blocking` helper as other fallible Tauri commands. Frontend callers still receive `Record<string, boolean>` on success.

## Test plan
- [x] `git diff --check`
- [x] `rustfmt --edition 2021 --check src/commands.rs`
- [x] `cargo check`
- [x] `cargo test -p acorn --lib` — 239 passed, 1 ignored
- [x] `pnpm run typecheck`
- [x] `pnpm run test` — 60 files passed, 578 tests passed, 1 skipped
- [x] `pnpm run build`
- [x] `pnpm run build:sidecar`

## Notes
- `pnpm run build` still emits existing Vite dynamic-import/chunk-size warnings, but exits successfully.

### DIFF
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -3040,10 +3040,18 @@ pub fn prepare_claude_fork(parent_uuid: String, new_cwd: String) -> AppResult<()
 /// the user right-clicks but absent from the last cache, or vice-versa.
 /// An on-demand scan locks the answer to "what's true right now."
 #[tauri::command]
-pub fn detect_session_agent(
+pub async fn detect_session_agent(
     state: State<'_, AppState>,
     session_id: String,
 ) -> AppResult<AgentDetection> {
+    let state = state.inner().clone();
+    run_blocking("detect_session_agent", move || {
+        detect_session_agent_inner(state, session_id)
+    })
+    .await
+}
+
+fn detect_session_agent_inner(state: AppState, session_id: String) -> AppResult<AgentDetection> {
     let parsed = parse_id(&session_id)?;
     let session_pids = crate::agent_resume_persister::collect_session_pids(&state);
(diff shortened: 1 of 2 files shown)


im-ian:

- `fix` label is applied.

Holding merge intentionally because the request was to prepare this PR for merge readiness, not to merge it.

PR #386, opened 2026-06-05
fix(terminal): keep plain space ASCII in WKWebView input

im-ian:
## Summary
This fixes intermittent terminal command failures where an unmodified Space can reach the shell as NBSP instead of an ASCII separator.

1. **Plain-space key path** — detect unmodified physical Space variants and send ASCII space directly before the xterm/WKWebView input path can surface NBSP.
2. **Modifier preservation** — leave Alt/Ctrl/Meta-space chords on the existing xterm path so intentional modified-space behavior is not rewritten.
3. **Regression coverage** — add terminal input helper tests and an IME E2E case that asserts NBSP-shaped plain Space writes ASCII space to the PTY.

## Expected impact
| Area | Before | After |
| --- | --- | --- |
| Terminal commands containing spaces | Intermittently `cd apps` / `pnpm i` could arrive as one NBSP-joined token and fail with `command not found`. | Plain physical Space is sent as ASCII `0x20`, so shells parse arguments normally. |
| Korean IME terminator space | Space could depend on the xterm/WKWebView input shape after the syllable flush. | The syllable still commits first, then ASCII space is sent explicitly. |
| Modified space and paste input | Existing xterm path. | Existing xterm path preserved for Alt/Ctrl/Meta-space and paste. |

## Design notes
- `isPlainSpaceKeydown` treats `code === "Space"`, legacy `Spacebar`, and NBSP key values as a physical space only when Alt/Ctrl/Meta are not held.
- The terminal keydown handler runs this after pending IME composition commits, so Korean terminator-space ordering stays intact.
- The direct write uses the same `sendUserInputToPty` path as ordinary terminal input, including multi-input fanout, and increments the terminal activity counter used by clipboard image fallback logic.

## Test plan
- [x] `pnpm run test -- src/lib/terminalInput.test.ts` — 61 files passed; 581 tests passed, 1 skipped
- [x] `pnpm run test:e2e -- tests/e2e/terminal-ime.spec.ts` — 11 passed
- [x] `pnpm run build` passes; Vite reports existing dynamic-import/chunk-size warnings

## Notes
- Local untracked `.codex-worktrees/` and `.codex/` directories were left unstaged.
- The Vite dynamic-import and chunk-size warnings are pre-existing build warnings, not caused by this PR.

### DIFF
--- a/src/components/Terminal.tsx
+++ b/src/components/Terminal.tsx
@@ -41,6 +41,7 @@ import {
   type ConversationNavigationDirection,
 } from "../lib/terminalConversation";
 import { patchTerminalMouseCoordinateScale } from "../lib/terminalMouseScale";
+import { isPlainSpaceKeydown } from "../lib/terminalInput";
 import { UI_SCALE_CHANGED_EVENT } from "../lib/layoutEvents";
 import {
   TERMINAL_PASTE_EVENT,
@@ -1063,6 +1064,10 @@ export function Terminal({
       }
     };
     let terminalActivityVersion = 0;
+    const sendKeyboardInputToPty = (data: string) => {
+      terminalActivityVersion += 1;
+      sendUserInputToPty(data);
+    };
     let imagePasteFallbackTimer: number | null = null;
     let imagePasteFallbackSerial = 0;
     const IMAGE_PASTE_FALLBACK_DELAY_MS = 500;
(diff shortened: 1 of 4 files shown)


jtf-ian:

- `fix` label is applied.

Holding merge intentionally because this ready-for-merge workflow stops before merging and leaves the final merge action to a human.

PR #396, opened 2026-06-08
fix(daemon): detach exited sessions from background registry

im-ian:
## Summary
This PR makes exited `acornd` PTY sessions detach from the daemon registry instead of lingering in Settings.

1. **Detach exited daemon sessions** — replace the daemon registry `mark_dead` path with generation-guarded detach logic so an exited PTY drops both its live handle and registry row.
2. **Preserve attached-stream exit status** — carry the PTY exit code on the live subscription so stream clients still receive an `Exit` frame with the code after the registry row detaches.
3. **Hide and clean inactive metadata** — keep Settings focused on live daemon sessions and add a bulk "Clear inactive sessions" action for stale metadata left by older daemon instances.
4. **Cover the UI path** — update daemon command mocks and add E2E coverage for hidden inactive rows plus the bulk cleanup action.

## Expected impact
| Area | Before | After |
| --- | --- | --- |
| Settings > Sessions | Exited daemon sessions could stay visible as inactive rows. | Only live daemon sessions are shown. |
| Daemon registry | Exited PTY metadata stayed resident until manually forgotten. | Normal PTY exit detaches the registry row. |
| Existing stale rows | Manual per-row cleanup only when visible. | One bulk cleanup button clears inactive metadata. |

## Design notes
- `PtyManager` now records the session generation (`pid` + `created_at`) before spawning the wait thread. The wait thread detaches only if that generation is still current, preventing an old wait thread from removing a newly respawned session with the same UUID.
- Exit codes moved onto the live PTY subscription path so active stream clients can still emit the correct `pty:exit` payload after daemon metadata detaches.
- `daemon_list_sessions` filters to live rows at the Tauri command boundary. The frontend also filters defensively so older daemon responses cannot surface inactive rows.
- `daemon_forget_inactive_sessions` exists as a migration/cleanup affordance for inactive rows produced by older daemon processes; live PTYs are left untouched.

## Test plan
- [x] `pnpm run build:sidecar` — passed; staged `acorn-ipc` and `acornd` sidecars.
- [x] `cargo check --manifest-path src-tauri/Cargo.toml` — passed.
- [x] `cargo test --manifest-path src-tauri/Cargo.toml -p acorn-daemon` — passed, 35 tests.
- [x] `PLAYWRIGHT_PORT=1437 pnpm exec playwright test tests/e2e/background-sessions.spec.ts` — passed, 3 tests.


im-ian:

- `fix` label is applied.

Holding merge intentionally because this workflow prepares the PR for merge readiness but does not merge.

PR #400, opened 2026-06-08
fix(session): hide internal title-generation prompts from History

im-ian:
## Summary
Acorn no longer exposes its own tab-title generation traffic as normal rows in Agents -> History.

1. **Title generation path** now tags Acorn's internal prompt and runs supported CLIs with their non-persistent flags where available: Codex uses `--ephemeral`, Claude uses `--no-session-persistence`.
2. **History parsing** now drops internal title-generation transcripts for Codex, Claude, and Antigravity so the rename workflow stays out of the normal session list.
3. **Graceful fallback** retries title generation with the unflagged command when the non-persistent flag is rejected, so older CLIs keep working while the prompt marker still hides any persisted transcript from History.
4. **Regression coverage** was added for prompt formatting, provider flags, the flag-rejection fallback, and History suppression across the supported agent paths.

## Expected impact
| Area | Before | After |
| --- | --- | --- |
| Agents -> History | Internal rename/title-generation prompts could show up as normal history rows | Those internal prompts are hidden |
| Title generation runtime | Codex / Claude could persist their title-generation sessions | Title generation runs with non-persistent flags on supported CLIs |
| Older CLI without the flag | Would have broken once a non-persistent flag was always appended | Falls back to the unflagged command; transcript still hidden via the marker |

## Design notes
- The implementation uses a stable internal prompt marker instead of trying to guess from prompt text alone.
- Codex and Claude have CLI-level non-persistence flags, so the generator can avoid writing session transcripts there.
- The flagged invocation falls back to the unflagged command on failure, so a CLI version that does not recognize the flag degrades gracefully instead of breaking title generation; the marker keeps the fallback transcript out of History.
- Antigravity does not expose an equivalent no-persistence flag in the CLI path we verified locally, so History filtering remains the containment point for that provider.

## Test plan
- [x] `cargo test -p acorn session_titles --manifest-path src-tauri/Cargo.toml`
- [x] `cargo test -p acorn agent_history --manifest-path src-tauri/Cargo.toml`
- [x] `git diff --check`
- [x] Local CLI verification for Codex and Claude: no new session-history files were created during title-generation runs
- [x] Fallback unit tests cover flag rejection (retry once unflagged), flag success (single run), and flag-less providers (no double-run on failure)

## Notes
- Antigravity still persists its own transcript files when run directly, based on local CLI verification. The PR hides the internal title-generation prompt from History rather than relying on a persistence-off flag there.

### DIFF
--- a/src-tauri/src/agent_history.rs
+++ b/src-tauri/src/agent_history.rs
@@ -293,6 +293,9 @@ fn parse_file_budget(limit: usize) -> usize {
 
 fn parse_codex_file(path: &Path, scope: HistoryScope<'_>) -> Option<AgentHistoryItem> {
     let state = parse_codex_state(path)?;
+    if state.internal_title_generation {
+        return None;
+    }
     let cwd = state.cwd.clone()?;
     if !scope.accepts_cwd(&cwd) {
         return None;
@@ -321,6 +324,9 @@ fn parse_codex_file(path: &Path, scope: HistoryScope<'_>) -> Option<AgentHistory
 
 fn parse_claude_file(path: &Path, scope: HistoryScope<'_>) -> Option<AgentHistoryItem> {
     let state = parse_claude_state(path)?;
+    if state.internal_title_generation {
+        return None;
+    }
     let cwd = state.cwd.clone()?;
     if !scope.accepts_cwd(&cwd) {
         return None;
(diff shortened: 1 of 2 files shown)


im-ian:

- `fix` label is applied.

Holding merge intentionally because this is a patch/fix PR and we want the pending SemVer patch/fix release to ship first without feature changes mixed into that patch line.

PR #423, opened 2026-06-10
refactor(ui): separate context submenu affordance

im-ian:
## Summary
Context menu trailing content now separates keyboard shortcut text from submenu affordances.

1. **Shared row renderer** — factors the common context-menu button row into a single internal renderer so button and submenu rows keep consistent spacing, hover, disabled, and accessibility behavior.
2. **Explicit trailing affordances** — renders submenu chevrons from the `type: "submenu"` path and leaves `shortcut` as keyboard hint text only.
3. **Regression coverage** — adds a focused ContextMenu test that locks submenu chevrons and shortcut hints onto separate rendering paths.

## Expected impact
| Area | Before | After |
| --- | --- | --- |
| Context menu model | Submenu affordance could be treated as special shortcut text | Submenu affordance is tied to `type: "submenu"` |
| Shortcut hints | Normal `<kbd>` hints rendered for shortcut strings | Unchanged for real shortcut strings |
| Menu row styling | Button and submenu rows duplicated similar JSX | Shared row renderer keeps row behavior aligned |

## Design notes
- This keeps the existing `ContextMenuItem` public shape intact: buttons use `shortcut?: string`; submenus use `children`.
- The renderer now has explicit `ContextMenuShortcut` and `ContextMenuSubmenuChevron` trailing components, which makes it harder to reintroduce sentinel shortcut values for non-shortcut UI.
- Existing submenu call sites in `Pane` and `Sidebar` already use `type: "submenu"`, so no caller migration is needed.

## Test plan
- [x] `pnpm exec vitest run src/components/ContextMenu.test.tsx` — passed, 4 tests
- [x] `pnpm run build` — passed; existing Vite dynamic-import/chunk-size warnings only

## Notes
- This PR intentionally follows up #422 by removing the `shortcut === ">"` special case rather than adding another call-site convention.

### DIFF
--- a/src/components/ContextMenu.test.tsx
+++ b/src/components/ContextMenu.test.tsx
@@ -66,6 +66,23 @@ describe("ContextMenu", () => {
     expect(onClick).toHaveBeenCalledTimes(1);
   });
 
+  it("renders submenu affordances separately from shortcut text", () => {
+    render([
+      {
+        type: "submenu",
+        label: "Copy",
+        children: [{ label: "Path", onClick: vi.fn() }],
+      },
+      { label: "Command palette", shortcut: "⌘K", onClick: vi.fn() },
+    ]);
+
+    const [copy, commandPalette] = document.querySelectorAll('[role="menuitem"]');
+    expect(copy?.textContent?.trim()).toBe("Copy");
+    expect(copy?.querySelector("svg")).not.toBeNull();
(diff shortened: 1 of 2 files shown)


im-ian:

- `refactor` label is applied.

Holding merge intentionally per the ready-for-merge workflow; this PR is waiting for human review/approval and was not merged by this workflow.

PR #447, opened 2026-06-15
fix(terminal): remove scrollback restore banner

im-ian:
## Summary
Remove the visible terminal restore banner while keeping scrollback replay behavior intact.

1. **No visible restore marker** — stop writing `— restored from previous session —` into terminal history after disk scrollback replay.
2. **Prompt spacing preserved** — keep a trailing newline so the newly spawned shell prompt does not overwrite the restored screen's final row.
3. **Legacy cleanup retained** — continue stripping previously saved restore markers from old scrollback snapshots before reuse or save.

## Expected impact
| Area | Before | After |
| --- | --- | --- |
| Terminal restart UX | Replayed scrollback ended with a visible restore message | Replayed scrollback returns silently with normal prompt spacing |
| Existing saved scrollback | Old restore marker lines could be present on disk | Old marker lines are still filtered out on load/save |

## Design notes
- The restore marker constant is no longer exported because the UI no longer writes the banner.
- `stripRestoreMarkers` still recognizes the old marker text so existing user data is cleaned opportunistically without a migration.
- The terminal replay path still serial-drains the restored bytes and reset sequences before spawning the PTY, preserving the cursor/input safety behavior around restored snapshots.

## Test plan
- [x] `pnpm run test -- src/lib/terminalScrollback.test.ts` passes — 64 files, 652 tests
- [x] `pnpm run build` passes; Vite reports existing chunk/dynamic-import warnings

### DIFF
--- a/src-tauri/crates/acorn-session/src/session.rs
+++ b/src-tauri/crates/acorn-session/src/session.rs
@@ -400,6 +400,21 @@ impl SessionStore {
         Ok(entry.clone())
     }
 
+    /// Flip the explicit auto-title opt-in. Used by the status poll to
+    /// promote a plain terminal session once an agent transcript is bound
+    /// to it. Does not bump `updated_at` — polling must not reshuffle the
+    /// sidebar's most-recent-first ordering. No-op when unchanged.
+    pub fn set_auto_title_enabled(&self, id: &Uuid, enabled: bool) -> SessionResult<Session> {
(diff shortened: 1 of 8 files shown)


im-ian:
Closing because this branch was based on a prior worktree commit and pulled unrelated changes into the PR. Reopening from a clean main-based branch.

PR #468, opened 2026-06-18
fix(diff): reset syntax state across hunks

im-ian:
## Summary
This PR fixes diff syntax highlighting when a displayed hunk contains an unterminated block comment.

1. **Hunk-scoped diff highlighting** — tokenize each unified-diff hunk independently so lexer state from a partial code range does not color later hunks as comments.
2. **Regression coverage** — add a TypeScript diff case where an open `/**` in one hunk must not make a later `export function` line render as a single comment token.

## Expected impact
| Area | Before | After |
| --- | --- | --- |
| Diff/code change view | An open block comment in one displayed hunk could make later hunks look commented out. | Syntax state resets at hunk boundaries, so later hunks highlight as code. |
| Full file code viewer | Unchanged. | Unchanged. |

## Design notes
- The public `highlightDiff` contract stays unchanged: callers still receive one highlighted HTML entry per parsed diff line, with `null` for hunk/meta/plain fallback lines.
- Context/add/delete lines are still split into new-side and old-side virtual sources inside each hunk, preserving side-specific highlighting without trusting lexer state across omitted unchanged ranges.
- The reset happens on `hunk` and `meta` lines because unified diffs do not include the full source between those boundaries.

## Test plan
- [x] `pnpm exec vitest run src/lib/highlight.test.ts` passes — 1 file / 3 tests
- [x] `pnpm run test` passes — 68 files / 702 tests

### DIFF
--- a/src/lib/highlight.test.ts
+++ b/src/lib/highlight.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, it } from "vitest";
-import { highlightThemeForMode } from "./highlight";
+import { parseDiff } from "./diff";
+import { highlightDiff, highlightThemeForMode } from "./highlight";
 
 describe("highlightThemeForMode", () => {
   it("uses a high-contrast light syntax theme for light app themes", () => {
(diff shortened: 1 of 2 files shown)


im-ian:

- `fix` label is applied.

Holding merge intentionally because `$acorn:ready-for-merge` prepares the PR for merge readiness and stops before merging.

PR #479, opened 2026-06-22
fix(terminal): normalize unicode spaces in user input

im-ian:
## Summary
Fixes a terminal input edge case where non-breaking-space-like characters could make shell commands arrive as a single word.

1. **Direct terminal input** — normalize Unicode space separators from browser `insertText` and captured `keypress` events before xterm emits them to the PTY.
2. **Paste input** — normalize the same shell-separator characters on the terminal paste text path, while keeping image paste fallback precedence unchanged.
3. **IME safety** — treat NBSP-like terminators as composition terminators so Korean IME commits remain single-write and do not duplicate syllables.
4. **Experimental escape hatch** — add a default-on Settings → Experiments toggle so users can preserve literal Unicode spaces in TUIs, REPLs, or parser tests when needed.
5. **Raw queued commands** — keep queued terminal commands/store state raw so non-interactive command sources are not globally rewritten.

## Expected impact
| Area | Before | After |
| --- | --- | --- |
| Terminal command input | NBSP-like separators could reach zsh as part of a single command word, e.g. `pnpm run dev` | User-entered NBSP-like separators are converted to ASCII spaces at input boundaries by default |
| Literal Unicode-space input | No dedicated product control for this edge case | Users can disable the experimental toggle to preserve NBSP-like characters in direct terminal input and paste |
| IME composition | NBSP-like terminators were not covered by the space terminator path | IME composition commits once, then emits a normal ASCII space while the toggle is enabled |
| PTY transparency | A broad `onData` rewrite would risk mutating all terminal data | Normalization is limited to user text input and paste boundaries |

## Design notes
- The PTY write path remains byte-transparent; `term.onData` still forwards xterm data without whitespace rewriting.
- The direct input handling happens before xterm's textarea listeners for `input` / `keypress`, matching the existing IME capture pattern in `Terminal.tsx`.
- `terminalPasteAction` takes the experiment setting as an input, so text paste normalization can be disabled without changing image-paste fallback behavior.
- The experiment defaults on for existing users and newly loaded settings, preserving the command-input fix unless a user explicitly opts out.
- `pendingTerminalInput` coverage asserts queued commands preserve Unicode spaces, preventing store-level normalization from creeping in later.

## Test plan
- [x] `pnpm exec vitest run src/lib/settings.test.ts src/lib/terminalPaste.test.ts src/lib/shellCommandWhitespace.test.ts` — 3 files / 84 tests passed
- [x] `pnpm run build` — passed; emitted existing Vite dynamic/static import and large chunk warnings
- [x] `pnpm exec playwright test tests/e2e/terminal.spec.ts -g "no-break"` — 4 tests passed
- [x] `pnpm exec playwright test tests/e2e/terminal-ime.spec.ts` — 11 tests passed
- [x] `git diff --check` / `git diff --cached --check` — passed

## Notes
- Build warnings about `api.ts` / `store.ts` dynamic imports and large chunks are pre-existing Vite warnings, not introduced by this PR.

### DIFF
--- a/src/components/SettingsModal.tsx
+++ b/src/components/SettingsModal.tsx
@@ -2489,6 +2489,20 @@ function ExperimentsSettings() {
         )}
       />
       <CheckboxRow
+        checked={experiments.normalizeTerminalUnicodeSpaces}
+        onChange={(checked) =>
+          patchExperiments({ normalizeTerminalUnicodeSpaces: checked })
+        }
+        label={st(
+          t,
+          "settings.experiments.normalizeTerminalUnicodeSpaces.label",
+        )}
+        description={st(
+          t,
+          "settings.experiments.normalizeTerminalUnicodeSpaces.description",
+        )}
+      />
+      <CheckboxRow
         checked={experiments.resumeModal}
         onChange={(checked) => patchExperiments({ resumeModal: checked })}
         label={st(t, "settings.experiments.resumeModal.label")}
--- a/src/components/Terminal.tsx
+++ b/src/components/Terminal.tsx
@@ -67,6 +67,7 @@ import {
   terminalPasteAction,
   type ClipboardImageFile,
 } from "../lib/terminalPaste";
+import { normalizeShellCommandWhitespace } from "../lib/shellCommandWhitespace";
 import { saveClipboardImageAttachment } from "../lib/clipboardImageAttachment";
 import {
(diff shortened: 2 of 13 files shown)


im-ian:

- `fix` label is applied.

Holding merge intentionally because `$acorn:ready-for-merge` prepares the PR for merge readiness without merging it.

PR #532, opened 2026-07-06
feat(workspace): rework kanban into a lifecycle board with PR/diff telemetry and quick-reply

im-ian:
## Summary
Rework the workspace kanban into a lifecycle board that surfaces what each session needs next.

1. **Lifecycle columns** — replace raw `SessionStatus` columns with derived `Idle → Working → Waiting → Review → Done` stages from `src/lib/kanbanLifecycle.ts`, including manual Done pins, PR-state handling, dirty-worktree review routing, dwell tracking, and stalled Working detection.
2. **PR and diff telemetry** — add `src/lib/useKanbanBoardData.ts` to poll PR listings per repo and git diff stats per session, with stale-request guards and error handling that avoids marking unknown diffs as clean.
3. **Kanban card controls** — add PR/detail chips, diff chips, dwell/stalled indicators, Waiting-stage quick reply, Mark as done / Restore from done, and preserve main's rename, current-PR, process metadata, and keyboard navigation behaviors.
4. **Persistence and coverage** — move board prefs to a stage-keyed v2 storage key, persist manual Done pins, and update unit/E2E coverage for lifecycle derivation, telemetry refresh, prefs, resize behavior, keyboard navigation, and PR metadata refresh.

## Expected impact
| Area | Before | After |
| --- | --- | --- |
| Board columns | Mirrored instantaneous backend status | Stable lifecycle stages for scanning work progress |
| Review signal | Hidden in branch/terminal context | Open PRs and dirty completed turns move cards into Review |
| Waiting sessions | Required opening terminal to respond | Waiting terminal cards accept inline quick replies |
| Card metadata | Session name, branch, worktree, recent message | Adds PR/check state, diff line counts, dwell time, stalled marker, process/current PR metadata |
| Saved board prefs | Status-keyed widths | Stage-keyed widths plus manual Done pins |

## Design notes
- Stage derivation is pure and frontend-only; no backend schema or persisted session status changes.
- The current status model is `ready`, `working`, `waiting_for_input`, and `errored`; `waiting_for_input` with a dirty worktree routes to Review, while `waiting_for_input`/`errored` otherwise remain Waiting.
- PR listing and diff polling use request sequencing so older responses cannot overwrite newer board context. Diff failures preserve the previous successful summary instead of falsely clearing Review signals.
- The kanban card root is a `div` with button semantics because Waiting cards contain a real input for quick reply; nested interactive content inside a `<button>` would be invalid and breaks focus behavior.
- Prefs intentionally move to `acorn:workspace-kanban:board-prefs:v2`; existing v1 column widths reset once because the column identity changed from status to stage.

## Follow-up (not in this PR)
- Drag-and-drop between lifecycle columns remains out of scope; manual movement is limited to Mark as done / Restore from done.
- The quick-reply path writes to the PTY directly. Richer command suggestions or per-agent reply templates should be separate work.

## Test plan
- [x] `pnpm exec tsc --noEmit` passed locally.
- [x] `pnpm exec vitest run src/lib/kanbanLifecycle.test.ts src/lib/useKanbanBoardData.test.ts src/lib/right-panel-cache.test.ts src/lib/kanbanBoard.test.ts` passed locally: 4 files, 67 tests.
- [x] `pnpm exec playwright test tests/e2e/kanban.spec.ts` passed locally: 16 tests.
- [x] `git diff --check` passed locally.
- [x] GitHub CI run 28998133519 passed: Frontend and E2E.

## Notes
- Users with existing kanban column widths will see default widths once after the v2 prefs migration.
- The branch was updated on top of current `main` after #578, #573, and #572 were merged.


im-ian:
Agent review record

Reviewed with a peer sub-agent plus a local diff pass. This PR is not merge-ready as reviewed.

Findings:

1. High: dirty finished turns do not reach `Review` for normal Claude/Codex flows. `deriveKanbanStage` returns `waiting` for every `needs_input` before considering dirty work, while turn completion commonly maps to `needs_input`, not `completed`. Result: finished dirty work can stay in `Waiting` with quick reply instead of moving to lifecycle `Review`. Add coverage for `needs_input + hasDiff` and decide whether that should transition to review.

2. High: diff telemetry can render misleading `+0 -0` for real changes. `useKanbanBoardData` sums `additions`/`deletions` from `fsGitStatus`, but that status API does not provide real line-count diff stats. Use real diff stats or render a changed-files indicator instead of line counts.

3. Medium: diff polling can apply stale async results out of order. Each interval starts a refresh and the resolving batch replaces the entire diff map without a sequence token or in-flight guard, so an older slow result can overwrite newer state and move cards between stages incorrectly.

4. Medium: missing-data paths can misclassify cards. PR fetch failures are swallowed without explicit stale/unknown state, so old PR data can keep cards in `Review`/`Done`; diff failures return clean, which can move dirty work back to `Idle`.

5. Medium: the current E2E failure is test drift from the old status-board model. The spec still expects `Needs input` and `data-kanban-column-status`, while the UI now renders lifecycle labels such as `Waiting` and `data-kanban-column-stage`. More assertions after the first failure use the old selectors too.

6. Medium: after PR detail merge/close, the board bumps `prRefreshKey`, but `rightPanelCache.fetchPullRequests(..., { force: true })` still reuses an existing in-flight request for the same key. A pre-mutation `all` listing can be reused after mutation and keep stale PR context until the next poll unless the PR cache is invalidated or force bypasses in-flight entries.

7. Low: hook/UI coverage is thin for the new board data. `useKanbanBoardData.test.ts` mostly covers PR selection, not polling races, error handling, PR/diff chips, quick reply, manual done, or lifecycle integration.

CI note: Frontend passes, E2E currently fails in `tests/e2e/kanban.spec.ts` on the stale `Needs input` heading assertion.

PR #544, opened 2026-07-06
fix(worktree): start new worktree sessions from default branch

im-ian:
## Summary
New worktree sessions now start from the project's stable default branch instead of inheriting whichever branch the project root currently has checked out.

1. **Default-branch base selection** — create Acorn-managed worktree branches from `main`, `origin/main`, `master`, or `origin/master` before falling back to current `HEAD`.
2. **Explicit worktree checkout** — create the branch first and pass it to libgit2 worktree creation, avoiding libgit2's implicit current-HEAD branch behavior.
3. **Regression coverage** — add a Rust test proving a feature-branch project root still creates a new worktree branch at `main`, while preserving the existing branch-name collision retry test.

## Expected impact
| Scenario | Before | After |
| --- | --- | --- |
| New worktree session from a project root currently on a feature/worktree branch | New branch started from that feature/worktree branch | New branch starts from the repo's default branch when available |
| Repos without `main`/`master` refs | Used current `HEAD` | Still falls back to current `HEAD` |
| Generated worktree branch name already exists | Retry suffix path | Retry suffix path preserved |

## Design notes
- The change stays in `worktree::create_worktree`, so all Acorn-managed worktree creation paths share the same default-branch policy.
- Remote-tracking refs (`origin/main`, `origin/master`) are accepted as base commits so fresh local checkouts without local default branches still get the expected starting point.
- If worktree creation fails after creating the new branch, the just-created branch is deleted before returning the error.

## Test plan
- [x] `git diff --check` — passed
- [x] `cargo test worktree::tests` — 16 passed
- [x] `cargo test create_unique_worktree_bumps_suffix_when_branch_name_taken` — passed

## Notes
- Playwright was not run locally; this is backend worktree creation behavior and is covered by focused Rust tests.

### DIFF
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -8110,19 +8110,15 @@ mod tests {
         repo
     }
 
-    // Regression: libgit2's default `Repository::worktree(name, path, None)`
-    // auto-creates a branch named after the worktree. When that branch
-    // already exists (e.g. the user's primary branch happens to share the
-    // repo basename), the call returned ErrorCode::Exists and bubbled up to
-    // the user as "failed to write reference 'refs/heads/<name>'". The
-    // retry loop now bumps the suffix on Exists just like it does for
-    // path collisions.
+    // Regression: a generated worktree branch name can collide with an
+    // existing branch. The retry loop bumps the suffix on Exists just like it
+    // does for path collisions.
     #[test]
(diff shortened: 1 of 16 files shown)


im-ian:
Superseded by #545, which carries the same fix on a clean main-based branch.

PR #572, opened 2026-07-09
feat(work-summary): show recent transcript messages

im-ian:
## Summary
This PR continues the local transcript recovery work by letting Work Summary show recent provider transcript messages, not only aggregate counts.

1. **Transcript summary payload** — `AgentTranscriptSummary` now includes a bounded `recent_messages` list with the latest user/assistant preview rows from the same parse pass that already computes message counts and token usage.
2. **Work Summary conversation detail** — terminal Work Summary tabs render recent transcript messages under the conversation metrics when provider JSONL metadata is available.
3. **Frontend contract coverage** — TypeScript types, component tests, and store mocks now cover the new summary field.

## Expected impact
| Area | Before | After |
| --- | --- | --- |
| xterm-limited conversation recovery | Work Summary showed message/token counts but not the actual latest transcript content | Work Summary can display recent local provider transcript messages without reading xterm history |
| Backend cost | Summary parsing counted messages and tokens only | The same pass keeps the latest 6 collapsed message previews, capped at 600 chars each |
| Existing chat/session behavior | Chat sessions and transcripts without messages showed the existing metrics/fallback UI | Existing fallback remains; recent message rows render only when summary metadata includes them |

## Design notes
- This is local-only: it reads already-discovered provider transcript files through existing summary commands and adds no remote-control behavior.
- The recent message list is deliberately bounded so large JSONL files do not inflate Tauri responses.
- The UI keeps messages inside the existing Conversation card rather than adding another top-level panel.

## Follow-up (not in this PR)
- Add a dedicated transcript drawer if users need longer history, search, or copy actions.
- Consider adding native chat recent-message previews to the same Work Summary section for parity.

## Test plan
- [x] `pnpm vitest run src/components/WorkSummaryView.test.tsx src/store.test.ts` — 158 passed
- [x] `cargo test --manifest-path src-tauri/Cargo.toml transcript_summary_counts_messages_and_uses_codex_cumulative_token_count` passes
- [x] `cargo check --manifest-path src-tauri/Cargo.toml` passes
- [x] `rustfmt --edition 2021 --check src-tauri/src/agent_history.rs` passes
- [x] `pnpm run test:e2e -- tests/e2e/work-summary.spec.ts` — 1 passed

## Notes
- This PR builds on the transcript path/provider metadata merged in PR #570, but it does not require PR #571.

### DIFF
--- a/src-tauri/src/agent_history.rs
+++ b/src-tauri/src/agent_history.rs
@@ -2201,6 +2201,64 @@ mod tests {
     }
 
     #[test]
+    fn transcript_summary_keeps_only_most_recent_messages() {
+        let dir = tempfile::tempdir().unwrap();
+        let transcript = dir
+            .path()
+            .join("rollout-2026-05-21T10-13-44-019e4818-7c15-7e60-9b3b-898a1c7803d6.jsonl");
+        let mut file = fs::File::create(&transcript).unwrap();
+        for index in 0..8 {
+            let role = if index % 2 == 0 { "user" } else { "assistant" };
+            writeln!(
+                file,
+                "{}",
+                serde_json::json!({
+                    "payload": {
+                        "role": role,
+                        "content": format!("message {index}"),
+                    },
+                })
+            )
+            .unwrap();
+        }
+        drop(file);
+
+        let item = AgentHistoryItem {
+            provider: AgentHistoryProvider::Codex,
+            id: "019e4818-7c15-7e60-9b3b-898a1c7803d6".to_string(),
+            title: "message 0".to_string(),
+            preview: Some("message 7".to_string()),
+            cwd: Some("/tmp/demo".to_string()),
+            worktree: None,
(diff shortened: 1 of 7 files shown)


im-ian:
Agent review record

Findings:

1. Medium: `RECENT_SUMMARY_MESSAGES` sliding-window behavior is not directly tested. Current fixtures stay under the cap, so there is no regression guard that more than six transcript messages evict the oldest while preserving order.

2. Low: `WorkSummaryView` can briefly show the previous session's recent messages during session switches because old chat/tokens/messages state is not reset while the next conversation is loading. This pattern existed before, but recent message text makes it more visible.

3. Low: recent message extraction does not dedupe consecutive duplicates the same way title context does. Also, `whitespace-pre-wrap` has little effect after `collapse_preview` single-lines text.

Suggested tests: Rust fixture with more than six qualifying messages, order/eviction assertion, and whitespace-only assistant text behavior.

PR #573, opened 2026-07-09
perf(terminal): batch output writes for hidden terminals

im-ian:
## Summary

This PR reduces terminal output rendering overhead without changing terminal behavior.

1. **Output batching:** Routes PTY output through a small `TerminalOutputWriter` that coalesces chunks before calling `xterm.write`.
2. **Hidden-terminal budget:** Keeps hidden terminal buffers current while using a slower/smaller inactive write cadence and skipping hidden DOM repaint work.
3. **Scrollback safety:** Normal scrollback saves wait briefly for queued output to drain before serializing, with a bounded timeout so shutdown is not held indefinitely.

## Expected impact

| Area | Before | After |
| --- | --- | --- |
| PTY output bursts | One `xterm.write` path per emitted chunk | Chunks are batched per frame for active terminals |
| Hidden terminals | Parsed output could still trigger viewport repaint work | Buffers stay updated, repaint work is reserved for active/on-screen terminals |
| Shutdown/save behavior | Scrollback serialized the current xterm buffer immediately | Saves give queued output a short drain window before serialization |

## Design notes

The change stays inside the existing TerminalHost/Terminal ownership model. It does not unsubscribe hidden terminals or drop output; TerminalHost still owns mount/eviction policy, and the new writer only changes how frontend output chunks are scheduled into xterm.

Inactive terminals use a conservative 80ms cadence and smaller batch budget. Active terminals flush on animation frames so interactive output still reaches the screen promptly.

## Follow-up (not in this PR)

Profile a real long-running agent session with several hidden terminals and tune the active/inactive byte budgets if traces show a better threshold.

## Test plan

- [x] `pnpm vitest run src/lib/terminalOutputWriter.test.ts src/lib/terminalRepaint.test.ts src/components/TerminalHost.test.tsx` -> 20 passed
- [x] `pnpm run typecheck` -> passed
- [x] `pnpm exec playwright test tests/e2e/terminal.spec.ts -g "does not write selected terminal text on right-click by default|keeps modifier-click link tooltip mounted while output streams"` -> 2 passed
- [x] `git diff --check` -> passed

### DIFF
--- a/src/components/Terminal.tsx
+++ b/src/components/Terminal.tsx
@@ -2447,18 +2447,8 @@ export function Terminal({
       isActive: () => isActiveRef.current,
     });
     outputWriterRef.current = outputWriter;
-    const OUTPUT_WRITER_IDLE_SAVE_TIMEOUT_MS = 200;
     const waitForOutputWriterIdleBeforeSave = async () => {
-      let timer: number | null = null;
-      await Promise.race([
-        outputWriter.whenIdle(),
-        new Promise<void>((resolve) => {
-          timer = window.setTimeout(resolve, OUTPUT_WRITER_IDLE_SAVE_TIMEOUT_MS);
-        }),
-      ]);
-      if (timer !== null) {
-        window.clearTimeout(timer);
-      }
+      await outputWriter.whenIdle();
     };
 
     // React StrictMode in dev runs effects twice (mount → cleanup → mount).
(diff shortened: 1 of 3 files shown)


im-ian:
Agent review record

Findings:

1. Medium: hidden/inactive terminal output batching has no total queued-byte cap. The flush budget limits write rate, but a high-output hidden terminal can still accumulate unbounded raw bytes in memory. Add backpressure, dropping policy, or a bounded queue with an explicit test.

2. Medium: the eviction/dispose path can drop queued output. If disposal forces a scrollback save while the writer is marked disposed, the idle wait is skipped and `outputWriter.dispose()` can clear batched-but-not-written bytes, so the persisted buffer may miss the tail.

3. Low/medium: debounced scrollback save can persist a partial buffer during large hidden bursts and may not save again after the writer drains unless another event occurs. Consider scheduling a save after drain/idle.

Suggested tests: first chunk larger than maxBytes, dispose mid-write, queue cap/backpressure behavior, inactive-to-active large backlog, and `whenIdle` with pending frame/timer but empty queue.

PR #580, opened 2026-07-09
fix(settings): expose pane focus shortcuts

im-ian:
## Summary
This fixes a Settings -> Shortcuts coverage gap for existing pane-focus hotkeys.

1. **Shortcut list coverage** — add the four pane-focus hotkeys to the Settings -> Shortcuts layout group so users can see, record, and reset them.
2. **Localized labels** — add English and Korean labels for focusing the pane left, right, above, and below.
3. **Regression coverage** — assert every `DEFAULT_HOTKEYS` id is present in the settings shortcut groups, and add an E2E check that a pane-focus shortcut can be recorded from Settings.

## Expected impact
| Area | Before | After |
| --- | --- | --- |
| Settings -> Shortcuts | Pane-focus shortcuts worked but could not be configured from Settings | Pane-focus shortcuts are visible and recordable |
| Future shortcut additions | Missing Settings rows could slip through | Component coverage fails if a configurable hotkey is omitted |

## Design notes
- The change keeps the existing manually grouped Settings layout and adds the missing pane-focus actions next to the other layout shortcuts.
- `SHORTCUT_GROUPS` is exported only so the component test can compare the rendered configuration against `HOTKEY_IDS` directly.

## Test plan
- [x] `pnpm exec vitest run src/components/SettingsModal.test.tsx` — 31 passed
- [x] `pnpm run test:e2e -- tests/e2e/settings.spec.ts -g "records a custom pane focus shortcut"` — settings spec ran, 12 passed
- [x] `pnpm run build` — passed

## Notes
- `pnpm run build` still emits existing Vite warnings about mixed static/dynamic imports and large chunks; this PR does not introduce those warnings.

### DIFF
--- a/src/components/SettingsModal.test.tsx
+++ b/src/components/SettingsModal.test.tsx
@@ -132,8 +132,9 @@ import {
   SESSION_TITLE_PROMPT_PREVIEW_MESSAGE,
   useSettings,
 } from "../lib/settings";
+import { HOTKEY_IDS } from "../lib/hotkeys";
 import { useAppStore } from "../store";
-import { SettingsModal } from "./SettingsModal";
+import { SettingsModal, SHORTCUT_GROUPS } from "./SettingsModal";
 
 const mockApi = vi.mocked(api);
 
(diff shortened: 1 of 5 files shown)


im-ian:
Agent review record

Findings: no issues found.

Residual risk: the new E2E verifies that the pane-focus shortcuts are listed and can be recorded, but it does not exercise the custom shortcut in a real multi-pane layout. Existing hotkey wiring looks consistent, so I would not block merge on that.

PR #653, opened 2026-07-14
fix(status): detect current Codex TUI turn starts

im-ian:
## Summary

Restore accurate working-state transitions for current Codex TUI sessions.

1. **Current turn-start detection** — recognize `from_tui` `op` records whose payload is a `UserTurn`, then forward them through the existing Acorn `start turn` hook channel.
2. **Backward compatibility** — retain the existing `codex_event/task_started` mapping for Codex versions that still emit the older event shape.
3. **Runtime regression coverage** — execute the generated wrapper against a representative current-format TUI record and assert that it emits the expected turn-start notification.

## Expected impact

| Scenario | Before | After |
| --- | --- | --- |
| Current Codex TUI begins a user turn | Session can remain `Idle` or `Waiting` while Codex is working | Session transitions to `Working` when the `UserTurn` record is written |
| Older Codex emits `task_started` | Turn start detected | Unchanged |
| Existing long-lived Codex process after app update | Keeps its already-running watcher | Requires a Codex process restart to load the refreshed wrapper |

## Design notes

- The change extends the existing generated-wrapper watcher instead of changing the hook server or session status contract.
- The match uses the current compact TUI recording shape observed in Codex session logs: `dir=from_tui`, `kind=op`, and `payload.UserTurn`.
- Completion handling remains unchanged: `agent-turn-complete` and `Stop` map to `needs_input`, which is rendered as `Waiting`.
- The test runs the generated shell wrapper rather than only asserting that its source contains a pattern, covering the event-to-notification behavior end to end.

## Follow-up (not in this PR)

- Evaluate first-class Codex lifecycle hooks for observing user-launched TUI sessions. Hook installation and trust UX should be designed separately.
- If Acorn later owns the Codex runtime, consider the Codex App Server `turn/started` and `turn/completed` notifications instead of parsing TUI recording output.

## Test plan

- [x] `cargo test --manifest-path src-tauri/Cargo.toml agent_wrappers::tests --lib` — 14 passed
- [x] `cargo test --manifest-path src-tauri/Cargo.toml --lib` — 384 passed, 1 ignored
- [x] `rustfmt --edition 2021 --check src-tauri/src/agent_wrappers.rs` — passed
- [x] `pnpm run test` — 95 files and 1,028 tests passed
- [x] `pnpm run build` — production build passed

## Notes

- Repository-wide `cargo fmt --all -- --check` still reports pre-existing formatting differences in unrelated Rust files. The changed file passes `rustfmt --check` independently.
- The frontend production build continues to report the existing chunk-size and mixed dynamic/static import warnings; the build succeeds and this Rust-only change does not affect those bundles.

### DIFF
--- a/src-tauri/src/agent_wrappers.rs
+++ b/src-tauri/src/agent_wrappers.rs
@@ -646,6 +646,53 @@ mod tests {
         }
     }
 
+    fn codex_wrapper_notifications_for_tui_line(line: &str) -> String {
+        let base = ScratchDir::new("codex-tui-event");
+        let wrapper_dir = ensure_agent_wrapper_dir_at(base.path()).unwrap();
+        let real_dir = base.path().join("real-bin");
+        fs::create_dir_all(&real_dir).unwrap();
+
+        let capture_path = base.path().join("notifications.log");
+        let session_log_path = base.path().join("session.jsonl");
+        write_executable(
+            &wrapper_dir.join("acorn-codex-notify"),
+            "#!/bin/sh\nprintf '%s %s\\n' \"$1\" \"$2\" >> \"$ACORN_NOTIFY_CAPTURE\"\n",
+        )
+        .unwrap();
+        write_executable(
+            &real_dir.join("codex"),
+            r#"#!/bin/sh
+: > "$CODEX_TUI_SESSION_LOG_PATH"
+sleep 0.1
+printf '%s\n' "$ACORN_TEST_TUI_LINE" >> "$CODEX_TUI_SESSION_LOG_PATH"
+_acorn_i=0
+while [ ! -s "$ACORN_NOTIFY_CAPTURE" ] && [ "$_acorn_i" -lt 50 ]; do
+  _acorn_i=$((_acorn_i + 1))
+  sleep 0.02
+done
+"#,
+        )
+        .unwrap();
+
+        let path = format!("{}:/usr/bin:/bin", real_dir.display());
+        let status = Command::new(wrapper_dir.join("codex"))
+            .env("PATH", path)
+            .env("ACORN_AGENT_WRAPPER_DIR", &wrapper_dir)
+            .env("ACORN_AGENT_HOOK_SESSION_ID", "session-1")
+            .env("ACORN_AGENT_HOOK_URL", "http://127.0.0.1:1/agent-hook")
+            .env("ACORN_AGENT_HOOK_TOKEN", "token-1")
+            .env("ACORN_NOTIFY_CAPTURE", &capture_path)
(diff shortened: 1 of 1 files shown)


im-ian:
## Investigation context

The investigated `full-prompt-delivery` session exposed two overlapping causes:

1. **The running Codex process predates the refreshed wrapper.** It started on July 13 at 17:45 and kept the watcher logic loaded at process start. Updating the staged wrapper does not hot-reload that already-running watcher, so an old session can prolong the incorrect status.
2. **The existing start-event matcher no longer matches current Codex TUI recordings.** Across 41 available TUI recording logs, none contained the previously expected `to_tui/codex_event/task_started` record. Thirty-four contained `from_tui/op/UserTurn`, including the most recent recordings. The wrapper therefore missed turn starts even though Codex was visibly `Working` and its rollout continued to advance.

The age of the session matters, but it is not the root cause by itself: a newly started session using the old matcher can still miss the current `UserTurn` shape.

## Why the status changed from `Idle` to `Waiting`

The lifecycle was asymmetric:

- Turn start was missed, leaving the persisted `ready` state rendered as `Idle`.
- Turn completion still reached Acorn through the configured `agent-turn-complete`/`Stop` notification.
- Acorn maps completion to `needs_input`, rendered as `Waiting`, because the agent is now awaiting another user instruction.

So `Idle → Waiting` means the end-of-turn callback was received; it does not mean start detection had repaired itself. Without this PR, a later active turn could remain displayed as `Waiting` while Codex was working.

## Fix in this PR

The generated Codex wrapper now treats the current `from_tui` + `op` + `payload.UserTurn` record as `start turn`, while retaining the older `codex_event/task_started` fallback. The regression test executes the actual generated wrapper and verifies the emitted notification.

Existing Codex processes must be restarted after installing an Acorn build containing this change. Restarting Acorn alone cannot replace watcher code already loaded inside a surviving Codex process.

## Longer-term lifecycle options

Manual TUI-record parsing should not be the final integration boundary:

- For observing a user-launched TUI, official [Codex hooks](https://learn.chatgpt.com/docs/hooks.md) expose lifecycle events such as `UserPromptSubmit`, `Stop`, and `PermissionRequest` with structured payloads.
- If Acorn owns the Codex runtime, the [Codex App Server](https://learn.chatgpt.com/docs/app-server.md) exposes explicit `turn/started` and `turn/completed` JSON-RPC notifications and can generate TypeScript/JSON schemas.
- Runtime-owned integrations can also use the [Codex SDK](https://learn.chatgpt.com/docs/codex-sdk.md).

This PR stays narrowly scoped to the current architecture. Automatically installing first-class hooks would also require hook trust/configuration UX, while moving to App Server or the SDK would change process ownership; both are better handled as separate design work.

PR #685, opened 2026-07-18
perf(terminal): move agent detection off the output hot path

im-ian:
## Summary
Move live agent detection out of the terminal output hot path while preserving image-paste compatibility.

1. **Remove output polling** — stop mounting a full agent scan and stop scheduling another scan for every 500 ms PTY output burst.
2. **Keep the compatibility path precise** — use the session's known provider as the fast path and run one fresh detection only when an image paste arrives without a provider hint.
3. **Lock the invoke contract down** — add a Playwright regression that proves streaming output performs zero detections while an unknown-provider image paste performs exactly one and still sends the agent Ctrl+V fallback.

## Expected impact

| Path | Before | After |
| --- | --- | --- |
| Busy terminal agent scans | Mount scan plus up to ~120 full scans/minute/pane during sustained output | 0 scans from mount or PTY output |
| Known-provider image paste | Background scans still run | 0 detection calls; provider hint is used directly |
| Unknown-provider image paste | Detection refreshed speculatively from output | 1 fresh detection at paste time |

## Design notes

- `detect_session_agent` deliberately refreshes the host process table and resolves live transcript mappings on every call, so it is appropriate for an explicit user action but too expensive for the PTY output path.
- Detection results are not cached across pastes. This avoids keeping a stale positive after an agent exits while the regular session-status path continues to provide the cheap known-provider hint.
- Context-menu agent detection remains unchanged; only the Terminal image-paste consumer moves to on-demand detection.

## Test plan

- [x] `PLAYWRIGHT_PORT=1420 node_modules/.bin/playwright test tests/e2e/terminal.spec.ts -g "detects image-paste agents on demand"` — 1 passed
- [x] `PLAYWRIGHT_PORT=1420 node_modules/.bin/playwright test tests/e2e/terminal.spec.ts` — 33 passed
- [x] `node node_modules/vitest/vitest.mjs run src/lib/terminalPaste.test.ts src/lib/agentProvider.test.ts` — 22 passed
- [x] `node node_modules/typescript/bin/tsc --noEmit` — passed
- [x] GitHub Actions required checks — Frontend and both E2E shards passed.

## Notes

- The regression test observed 3 `detect_session_agent` calls on the old path after mount plus two output chunks; the optimized path records 0 before the image paste and exactly 1 afterward.

### DIFF
--- a/src/components/Terminal.tsx
+++ b/src/components/Terminal.tsx
@@ -972,36 +972,8 @@ export function Terminal({
     const unlistenFns: UnlistenFn[] = [];
     let observedLinkedWorktreePath: string | null = null;
     let liveCwdProbeTimer: number | null = null;
-    let agentProbeTimer: number | null = null;
     let daemonSessionAliveAtMount = false;
     let replayScrollbackOnSpawn = true;
-    let agentImagePasteFallbackActive = providerSupportsImagePasteFallback(
-      pasteAgentProviderRef.current,
-    );
-
-    const refreshAgentDetection = () => {
-      void api
-        .detectSessionAgent(sessionId)
-        .then((agent) => {
-          if (disposed) return;
-          agentImagePasteFallbackActive =
-            providerSupportsImagePasteFallback(pasteAgentProviderRef.current) ||
-            detectionHasImagePasteFallbackProvider(agent);
-        })
-        .catch((err: unknown) => {
-          console.debug("[Terminal] agent detection failed", err);
-        });
-    };
-
-    const scheduleAgentDetection = () => {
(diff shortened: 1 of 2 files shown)


im-ian:
## Merge readiness

- CI: Frontend and both E2E shards passed.
- Mergeability: `CLEAN` / `MERGEABLE`.
- Changelog label: exactly one `perf` label.
- Local verification: terminal Playwright suite (33 tests), focused Vitest suite (22 tests), and TypeScript typecheck passed.

Holding merge intentionally because this task asks for separate reviewable PRs; merging remains a separate user decision.

PR #686, opened 2026-07-18
perf(bundle): ship only Acorn's supported Shiki grammars

im-ian:
## Summary

Replace Shiki's full language/theme bundle with a typed, Acorn-specific registry so production builds contain only syntax assets the app can select.

1. **Use fine-grained Shiki packages** — replace the root `shiki` dependency with `@shikijs/core`, the existing JavaScript regex engine, and explicit language/theme packages.
2. **Bound the lazy registry** — keep the existing 39 language mappings and two app themes, but give Vite one dynamic import per supported registration instead of exposing Shiki's complete catalog.
3. **Guard the bundle contract** — add tests that lock the supported registry and verify representative language/theme registrations resolve.

## Expected impact

| Production frontend metric | `main` | This PR | Change |
| --- | ---: | ---: | ---: |
| Total `dist` bytes | 12,464,859 B | 5,679,870 B | **-6,784,989 B (-54.43%)** |
| Aggregate gzip bytes | 3,032,036 B | 1,590,899 B | **-1,441,137 B (-47.53%)** |
| Emitted files | 307 | 47 | **-260 (-84.69%)** |
| Entry JS bytes | 2,416,609 B | 2,374,935 B | **-41,674 B (-1.72%)** |
| Entry JS gzip bytes | 696,984 B | 681,108 B | **-15,876 B (-2.28%)** |

Measurements use clean production builds from the same checkout and aggregate every emitted file; gzip figures use `gzip -9` for a consistent before/after comparison.

## Design notes

- `langFromPath` keeps the same extension, filename, and fence-language behavior; its return type is now derived from the registry, so a mapping cannot reference a grammar that is not bundled.
- Languages and themes remain lazy chunks. The change narrows the available loader table rather than eagerly loading all 39 grammars.
- The pure-JavaScript regex engine and the JavaScriptCore anchor workaround are unchanged, preserving the hardened WKWebView CSP behavior.
- Unknown file types still fall back to plain text before Shiki is invoked.

## Follow-up (not in this PR)

- Lazy-mounting Settings and Command Palette can reduce the initial entry chunk further, but does not reduce total shipped bytes and should remain a separate startup-focused change.
- The Acorn illustration can be resized to its rendered resolution for another asset reduction after cross-platform visual QA.

## Test plan

- [x] `pnpm install --frozen-lockfile --offline` — lockfile accepted with the fine-grained dependencies.
- [x] `pnpm exec vitest run src/lib/highlightRegistry.test.ts src/lib/highlight.test.ts` — 8 tests passed.
- [x] `pnpm test` — 100 files / 1,125 tests passed.
- [x] `pnpm typecheck` — passed.
- [x] `pnpm build` — TypeScript and Vite production build passed; output measured at 5,679,870 B across 47 files.
- [x] GitHub Actions CI — Frontend and both E2E shards passed.

## Notes

- Dependabot PR #683 also touches `package.json` and `pnpm-lock.yaml`; whichever PR lands second may need to regenerate its lockfile.
- Vite still reports the pre-existing mixed static/dynamic import notices and the large entry/CPP chunk warning. Those warnings are not introduced by this PR; the measured entry and total output are both smaller.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,10 @@
     "@dnd-kit/core": "^6.3.1",
     "@dnd-kit/sortable": "^10.0.0",
     "@dnd-kit/utilities": "^3.2.2",
+    "@shikijs/core": "^4.0.2",
+    "@shikijs/engine-javascript": "^4.0.2",
+    "@shikijs/langs": "^4.0.2",
+    "@shikijs/themes": "^4.0.2",
     "@tanstack/react-virtual": "^3.13.24",
     "@tauri-apps/api": "^2.11.1",
     "@tauri-apps/plugin-dialog": "^2.7.1",
@@ -57,7 +61,6 @@
     "rehype-raw": "^7.0.0",
     "rehype-sanitize": "^6.0.0",
     "remark-gfm": "^4.0.1",
-    "shiki": "^4.0.2",
     "tailwind-merge": "^3.5.0",
     "tinykeys": "^3.0.0",
     "zustand": "^5.0.14"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -17,6 +17,18 @@ importers:
       '@dnd-kit/utilities':
         specifier: ^3.2.2
         version: 3.2.2(react@19.2.7)
+      '@shikijs/core':
+        specifier: ^4.0.2
+        version: 4.0.2
+      '@shikijs/engine-javascript':
+        specifier: ^4.0.2
+        version: 4.0.2
+      '@shikijs/langs':
+        specifier: ^4.0.2
+        version: 4.0.2
+      '@shikijs/themes':
+        specifier: ^4.0.2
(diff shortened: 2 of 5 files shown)


im-ian:
## Merge readiness

- CI: Frontend and both E2E shards passed.
- Mergeability: `CLEAN` / `MERGEABLE`.
- Changelog label: exactly one `perf` label.
- Local verification: 1,125 Vitest tests, TypeScript typecheck, frozen lockfile install, and production build passed.
- Independent review: no blockers; all 39 language loaders and both themes were runtime-checked against Shiki 4.0.2.
- Measured output: 12,464,859 B → 5,679,870 B (`-54.43%`); aggregate gzip 3,032,036 B → 1,590,899 B (`-47.53%`).

Holding merge intentionally because this task asks for separate reviewable PRs; merging remains a separate user decision.

PR #745, opened 2026-08-04
feat(sidebar): convert workspaces to projects and back

im-ian:
## Summary

Adds two sidebar context-menu conversions between the workspace ("source folder") and project levels, so a directory can move between the two groupings without losing its sessions.

- **Right-click a worktree workspace title → "Convert to project".** The workspace directory is registered as a standalone project and every session in that workspace is re-homed under it.
- **Right-click a project title → "Convert to workspace of …".** The project folds back in as a workspace of a host project that physically contains it; its own worktree workspaces are recreated under the host so nothing is orphaned.

## How it works

A project is a repository root and a workspace is a directory inside one, so the conversions are only offered where the paths actually nest that way:

- Promote is shown only for workspaces whose directory differs from the project root, and the backend additionally rejects anything that is not a repository root of its own — a plain subdirectory resolves back to the repo it already belongs to and could never hold sessions.
- Fold lists only projects that contain the source project's path, which covers Acorn-managed worktrees (`.acorn/worktrees/*`) and monorepo subdirectories.

Two new commands back this:

- `open_project_at_path` registers a directory that already belongs to a registered project. There is no folder picker, so unlike `add_project` it creates no new trust root: the path must already be reachable from a project Acorn knows about.
- `move_sessions_to_project` re-homes sessions, validating every session against the target project before any of them moves, using the same rule `create_session` applies to a new session's cwd.

`SessionStore::update_repo_path` clears `isolated` when the session's worktree becomes the new project root, so worktree-deletion prompts never offer to delete the project root itself. The on-disk `worktree_path` is untouched, so a live PTY keeps its cwd across the conversion.

## Test plan

- [x] `pnpm run typecheck`
- [x] `pnpm run test` — 1232 passed (5 new store tests covering promote, fold, and both refusal guards)
- [x] `pnpm run build:frontend`
- [x] `cargo test --lib` — 659 passed (1 new `update_repo_path` test)
- [ ] Manual run of the desktop app — not executed in this environment

## Notes

Detecting git worktrees kept outside the repo tree is intentionally left out: the fold target list uses a path-prefix check only. Worth adding a real worktree lookup if someone keeps worktrees elsewhere.

### DIFF
--- a/src-tauri/crates/acorn-session/src/session.rs
+++ b/src-tauri/crates/acorn-session/src/session.rs
@@ -1351,6 +1351,29 @@ impl SessionStore {
         Ok(entry.clone())
     }
 
+    /// Re-home a session under a different registered project. Used when a
+    /// workspace folder is promoted to a standalone project (and the reverse:
+    /// a project folded back in as a workspace of another project). Only the
+    /// project association moves — the on-disk `worktree_path` is untouched,
+    /// so the live PTY keeps its cwd. `isolated` is cleared when the session's
+    /// worktree becomes the new project root: it is no longer a throwaway
+    /// worktree *inside* a project, and worktree-deletion prompts must not
+    /// offer to delete the project root itself.
+    pub fn update_repo_path(
+        &self,
+        id: &Uuid,
+        repo_path: std::path::PathBuf,
+    ) -> SessionResult<Session> {
+        let mut entry = self
+            .inner
+            .get_mut(id)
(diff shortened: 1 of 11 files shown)


jtf-ian:
Superseded: this targeted workspace folders, but the request is about source folders (#738). Redoing on top of current main.

im-ian:
Superseded: this targeted workspace folders, but the request is about source folders (#738). Redoing on top of current main.

PR #783, opened 2026-08-12
fix(scrollback): preserve history on permission errors

im-ian:
## Summary

Preserve persisted terminal history when Acorn cannot read it, instead of treating the failure as an empty snapshot that is safe to overwrite.

1. **Distinguish absence from access failure.** `scrollback::load` and `scrollback::delete` now treat only `NotFound` as an idempotent empty/missing state; permission and other I/O errors reach the caller.
2. **Fail closed in the terminal.** Scrollback saves remain disabled unless the initial snapshot is read and applied successfully. The PTY still starts, and the terminal shows an actionable warning that saving is disabled to protect existing data.
3. **Cover the data-loss boundary.** Rust tests exercise real Unix permission denial and non-missing read failures, while Playwright verifies that live PTY output continues without invoking `scrollback_save` after a denied load.

## Expected impact

| Scenario | Before | After |
| --- | --- | --- |
| Existing scrollback is unreadable | Reported as no snapshot; later output could atomically replace it | Error is preserved; saving stays disabled and the existing file is left untouched |
| Scrollback file is genuinely missing | Empty terminal baseline | Same empty terminal baseline; future saves remain enabled |
| Terminal startup after a load error | PTY starts silently | PTY still starts, with an in-terminal preservation warning |
| Delete encounters an inaccessible path | `Path::exists()` could collapse the error into success | Permission and other non-`NotFound` failures are returned |

## Design notes

- The backend contract now makes `Ok(None)` mean exactly “no persisted file,” so the frontend can safely use successful invocation as the write-authorization boundary.
- Saving stays disabled for that terminal mount after any load/apply failure. This intentionally trades persistence of new output for preservation of a snapshot Acorn never successfully read; remounting after permissions are repaired retries the load normally.
- The PTY lifecycle is independent of persistence recovery, so a scrollback filesystem problem does not prevent the shell from remaining usable.

## Test plan

- [x] `cargo test --manifest-path src-tauri/Cargo.toml -p acorn-session scrollback::tests -- --test-threads=1` — 9 passed.
- [x] `cargo test --manifest-path src-tauri/Cargo.toml --workspace --locked -- --test-threads=1` — 1,045 passed, 1 existing ignored.
- [x] `pnpm test` — 112 files / 1,350 tests passed.
- [x] `pnpm typecheck` — passed.
- [x] `pnpm exec playwright test tests/e2e/terminal.spec.ts` — 44 passed.
- [x] `pnpm build` — production bundle completed.
- [x] `cargo clippy --manifest-path src-tauri/Cargo.toml -p acorn-session --all-targets --locked` — completed with no warnings from the changed scrollback code.

## Notes

- A stricter workspace-wide Clippy run with `-D warnings` remains blocked by existing warnings in `acorn-pty`, `acorn-transcript`, and `acorn-daemon`; none are introduced by or located in this change.
- The Vite production build retains the repository's existing chunk-size and ineffective-dynamic-import warnings; the build succeeds.

### DIFF
--- a/src-tauri/crates/acorn-session/src/scrollback.rs
+++ b/src-tauri/crates/acorn-session/src/scrollback.rs
@@ -6,8 +6,9 @@
 //! mount it loads via `scrollback_load` and `term.write`s the bytes back into
 //! xterm before spawning the PTY.
 //!
-//! Atomic writes use same-directory replace semantics. Files are best-effort: read errors
-//! return `None` so the UI can proceed with an empty buffer.
+//! Atomic writes use same-directory replace semantics. A missing file is an
+//! empty buffer, but other read failures remain errors so callers cannot
+//! mistake an inaccessible snapshot for an empty one and overwrite it.
 //!
 //! Callers pass the application's data directory in explicitly so this crate
 //! does not depend on the main `acorn` crate's `persistence::data_dir()`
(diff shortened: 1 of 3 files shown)


im-ian:

This PR made an inaccessible scrollback snapshot stop being mistaken for an empty one, and kept saves disabled after a failed load so the unread snapshot could not be overwritten. `main` now does all of that:

- `scrollback.rs` — `NotFound` returns `Ok(None)`; every other read failure is an error. `main` reads through `acorn_platform::fs::open_regular_nofollow` with a `MAX_PAYLOAD_BYTES` bound, which is stricter than the `fs::read_to_string` proposed here (no symlink follow, no unbounded read).
- `Terminal.tsx` — the save gate flips to `true` only after a successful initial load, so a failed load leaves it closed for the lifetime of the mount and later PTY output cannot clobber the snapshot. The debounced-save path skips while loading or after an error for the same reason.
- the failure is surfaced through `showTranslatedErrorToast("toasts.session.scrollbackLoadFailed", err)`.
- `tests/e2e/terminal.spec.ts` already carries the equivalent case, `"scrollback load failure keeps later terminal saves disabled"`.

Rebasing this branch would have traded the bounded no-follow read for `fs::read_to_string` and replaced the toast with an in-terminal message, with no behavioural gain, so it is closed rather than landed.

PR #784, opened 2026-08-12
fix(persistence): preserve inaccessible state-file entries

im-ian:
## Summary

Make optional persistence reads distinguish a truly absent file from an inaccessible or otherwise occupied state-file path.

1. **Define one optional-read contract.** Add `acorn_platform::fs::read_optional`, where `None` means no filesystem entry exists and permission errors, invalid file types, and dangling symlinks remain errors.
2. **Protect boot cleanliness decisions.** Session and project loaders now mark inaccessible or dangling state files unclean instead of reporting an authoritative empty store, preserving the existing pane-reconciliation and scrollback-prune guards.
3. **Stop per-record replacement after failed reads.** Project settings, chat session state, and Graph run state propagate optional-read failures before their read-modify-write flows can publish a replacement.
4. **Exercise real failure modes.** Regression tests cover Unix permission denial and dangling symlinks, including assertions that chat and project-settings mutations leave the occupied path intact.

## Expected impact

| Scenario | Before | After |
| --- | --- | --- |
| State-file parent is inaccessible | `Path::exists()` could report false and classify the store as cleanly empty | Load is unclean and destructive empty-state reconciliation stays guarded |
| Chat or project-settings path is a dangling symlink | Treated as new state and atomically replaced during mutation | Mutation fails before writing and the symlink remains intact |
| Optional state file is genuinely absent | Empty/default state | Same empty/default state |
| Existing readable file | Parsed normally | Same parsed state |

## Design notes

- The helper reads first, then uses `symlink_metadata` only to disambiguate a `NotFound` read. This avoids a separate positive-existence precheck while still recognizing a dangling symlink as an occupied path.
- A race that creates an entry between the read and metadata check fails closed for that attempt; the caller can retry rather than replace state it did not read.
- Per-record chat and project-settings mutations already load before saving, so propagating the read error is sufficient to stop replacement in those flows.

## Follow-up (not in this PR)

- Session and project stores still need the separately proposed process-wide write fence when boot loading is unclean. This PR makes their cleanliness signal authoritative but intentionally does not introduce the broader read-only-mode behavior or touch every mutation surface.
- The remembered project-parent marker has different recovery semantics because its caller currently clears both malformed and inaccessible values; it will be handled as a separate work unit.

## Test plan

- [x] `cargo test --manifest-path src-tauri/Cargo.toml -p acorn-platform fs::tests -- --test-threads=1` — 5 passed.
- [x] `cargo test --manifest-path src-tauri/Cargo.toml --lib persistence::tests -- --test-threads=1` — 21 passed.
- [x] `cargo test --manifest-path src-tauri/Cargo.toml --lib project_settings::tests -- --test-threads=1` — 6 passed.
- [x] `cargo test --manifest-path src-tauri/Cargo.toml --workspace --locked -- --test-threads=1` — 1,049 passed, 1 existing ignored.
- [x] `cargo clippy --manifest-path src-tauri/Cargo.toml -p acorn-platform -p acorn --lib --tests --locked` — completed; no warning points to the changed optional-read code.
- [x] `rustfmt --edition 2021 src-tauri/crates/acorn-platform/src/fs.rs src-tauri/src/persistence.rs src-tauri/src/project_settings.rs` and `git diff --check` — clean.

## Notes

- Repository-wide `cargo fmt --all --check` remains blocked by a pre-existing formatting diff in `src-tauri/src/pty_env.rs`; that unrelated file is not changed here.
- Clippy continues to report the repository's existing warnings in unrelated modules and pre-existing tests; this change introduces no new Clippy diagnostic.

### DIFF
--- a/src-tauri/crates/acorn-platform/src/fs.rs
+++ b/src-tauri/crates/acorn-platform/src/fs.rs
@@ -6,6 +6,26 @@ use std::path::Path;
 
 use tempfile::NamedTempFile;
 
+/// Read an optional file without collapsing filesystem errors into absence.
+///
+/// `None` means no directory entry exists at `path`. In particular, a
+/// dangling symlink remains an error even though following it produces
+/// `NotFound`; callers must not treat an occupied but unreadable state-file
+/// path as a fresh file that is safe to replace.
+pub fn read_optional(path: &Path) -> io::Result<Option<Vec<u8>>> {
+    match fs::read(path) {
+        Ok(bytes) => Ok(Some(bytes)),
+        Err(read_error) if read_error.kind() == io::ErrorKind::NotFound => {
+            match fs::symlink_metadata(path) {
+                Ok(_) => Err(read_error),
+                Err(metadata_error) if metadata_error.kind() == io::ErrorKind::NotFound => Ok(None),
+                Err(metadata_error) => Err(metadata_error),
+            }
+        }
+        Err(error) => Err(error),
+    }
+}
+
 /// Atomically replace `path` with `contents` after flushing it to disk.
 pub fn write_atomic(path: &Path, contents: &[u8]) -> io::Result<()> {
     write_atomic_with_mode(path, contents, false)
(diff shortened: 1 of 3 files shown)


im-ian:

This PR distinguished "state file is absent" from "state file could not be read", so an unreadable `sessions.json` / `projects.json` / project settings file would not be treated as an empty one and overwritten. `main` now carries that end to end:

- `read_bounded_regular_file` separates the arms: `NotFound` → start empty and report **clean**; any other read failure → start empty but report **not clean**.
- `AppState::sessions_loaded_cleanly` / `projects_loaded_cleanly` carry that signal, and the `load_status` command exposes it. Per its doc comment, the frontend "skips the pane-wipe code path so a transient disk glitch does not erase the persisted layout".
- `project_settings::load_all` propagates any non-`NotFound` error instead of returning an empty map.

`main`'s reader is also stricter than the `acorn_platform::fs::read_optional` proposed here: it enforces the per-file `MAX_*_FILE_BYTES` limit and re-checks `dev`/`ino` after opening. `backup_corrupt_file(&path, &bytes)` on `main` likewise backs up the bytes already read rather than re-reading the path.

Rebasing this branch would have traded those protections for no behavioural gain, so it is closed rather than landed.

PR #789, opened 2026-08-12
fix(daemon): make PID locking permission-safe

im-ian:
## Summary

This makes the daemon singleton claim fail closed on filesystem access errors and removes PID-file startup races.

1. **Exclusive singleton ownership.** Replace the exists/read/write sequence with an OS file lock held by a process-lifetime guard, so concurrent daemon starts cannot both acquire the claim.
2. **Fail-closed file handling.** Propagate PID-file metadata, open, read, and write failures instead of treating unreadable state as stale; reject symlinks, non-regular files, and Unix hard links without touching their targets.
3. **Stable lifecycle semantics.** Reuse the lock inode across daemon runs, tighten Unix permissions to 0600, and release ownership by dropping the guard rather than unlinking a path another process could recreate.
4. **Cross-platform contention.** Treat the OS lock as authoritative and make the diagnostic PID optional because Windows denies reads through a second handle while the lock is held.
5. **Regression coverage.** Cover simultaneous acquisition, stale and recycled PID recovery, permission denial, link rejection, persisted lockfiles, owner-only modes, and platform-specific contended-read behavior.

## Expected impact

| Area | Before | After |
| --- | --- | --- |
| Concurrent startup | Check-then-write race could let multiple processes claim startup | OS locking serializes every contender |
| Access errors | Unreadable claims were treated as stale and could be overwritten | Startup fails with the original filesystem error |
| Windows contention | A second daemon failed with a raw locked-range read error | It reports that a daemon is already running; PID detail is omitted when Windows withholds it |
| Path safety | Writes followed links and graceful cleanup unlinked by pathname | Link-shaped claims are rejected and the locked inode is retained |
| Lockfile privacy | Existing modes were preserved and new modes followed defaults | Unix lockfiles are normalized to 0600 |

## Design notes

- The held OS lock is authoritative for daemons using this implementation. The PID and executable check remains after lock acquisition to avoid starting beside a live legacy daemon that wrote a PID but did not hold an OS lock.
- Reading the PID after a contended lock is diagnostic only. Unix normally permits it; Windows returns the same locked-range error that proves the other owner is active, so the public outcome carries `Option<u32>`.
- The PID file intentionally remains after shutdown. Its presence is no longer a liveness signal; keeping the same inode avoids an unlock/unlink/recreate window during handoff.
- Socket cleanup remains independent and still occurs before the PID guard is released.

## Test plan

- [x] `cargo test --manifest-path src-tauri/Cargo.toml -p acorn-daemon` — 51 passed after the Windows contract adjustment.
- [x] `cargo test --workspace -- --test-threads=1` — full Rust workspace passed before the Windows-only follow-up.
- [x] `cargo check --manifest-path src-tauri/Cargo.toml --bin acornd` — passed.
- [x] `cargo check --manifest-path src-tauri/Cargo.toml -p acorn-daemon --target x86_64-pc-windows-msvc` — passed.
- [x] `cargo clippy -p acorn-daemon --lib` — passed with existing warnings only.
- [x] `git diff --check` — passed.

## Notes

- A macOS cross-check of the complete Windows `acornd` binary is blocked by missing MSVC C headers for native dependencies; the daemon crate itself cross-checks successfully and CI runs the native Windows test binary.
- `cargo clippy --workspace --all-targets -- -D warnings` is currently blocked by pre-existing warnings in acorn-pty, acorn-transcript, acorn-ipc, and unchanged daemon files. None point to the PID-lock implementation in this PR.

### DIFF
--- a/src-tauri/crates/acorn-daemon/src/lifecycle.rs
+++ b/src-tauri/crates/acorn-daemon/src/lifecycle.rs
@@ -22,8 +22,9 @@
 //!    a short timeout; on success the daemon is alive, on EOF / refused
 //!    the slot is free.
 
-use std::io;
-use std::path::PathBuf;
+use std::fs::{File, Metadata, OpenOptions, TryLockError};
+use std::io::{self, Read, Seek, Write};
+use std::path::{Path, PathBuf};
 
 use sysinfo::{Pid, ProcessRefreshKind, ProcessesToUpdate, RefreshKind, System};
 
@@ -36,10 +37,24 @@ const DAEMON_EXECUTABLE: &str = "acornd";
 /// Outcome of attempting to acquire the daemon singleton lock.
 #[derive(Debug)]
 pub enum PidLock {
-    /// We hold the lock. The file now contains our PID.
-    Acquired(PathBuf),
+    /// We hold the lock. The guard keeps the OS lock alive until it is
+    /// dropped; the reusable PID file itself remains on disk.
+    Acquired(PidLockGuard),
     /// Another daemon is already running. Field is its PID.
     AlreadyHeld(u32),
+}
+
+/// Process-lifetime ownership of the daemon PID file.
+#[derive(Debug)]
+pub struct PidLockGuard {
+    path: PathBuf,
+    _file: File,
+}
+
+impl PidLockGuard {
+    pub fn path(&self) -> &Path {
+        &self.path
+    }
 }
 
 /// Attempt to acquire the singleton lock. Returns immediately — no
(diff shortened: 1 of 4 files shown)


im-ian:

This PR reworked the daemon PID lock so permission and metadata failures could not be mistaken for "no daemon running". `main`'s `lifecycle.rs` now does that, and goes further:

- `read_pid_file_at` opens through `acorn_platform::fs::open_regular_nofollow`, so a symlinked or special pid file is refused instead of followed.
- the PID read is bounded by `MAX_PID_FILE_BYTES`, with an oversized file treated as absent rather than read unbounded.
- the lock file is forced to `0600` and rejected outright if it is not a regular file.
- `pid_file_lock_conflict` separates "another process holds the lock" from a real I/O failure; only the former degrades to `Ok(None)`, everything else propagates.
- `PidLockGuard` clears the PID while still holding the lock, keeps the inode in place so release cannot race a concurrent opener, and releases through `impl Drop`.

That last point also makes the `acornd.rs` half of this branch unnecessary: the explicit `drop(pid_lock)` before returning a `bind_both` error is what `Drop` already does when the guard leaves scope.

Rebasing this branch would have replaced the hardened implementation with a parallel one built on `TryLockError` + `sysinfo`, so it is closed rather than landed.

PR #791, opened 2026-08-12
fix(transcripts): surface access failures in title and resume flows

im-ian:
## Summary — first slice only

This branch is now **just the `session_titles` slice** of the original 1,040-line change. The rest follows separately; see below.

`resolve_chat_title_input` ended in `.ok()?`, so every failure became "nothing to title". `load_chat_session_state` already maps a missing file to an empty state, so the only failures reaching that `.ok()` were real: the state file is unreadable, corrupt, or belongs to another session.

The caller turns `None` into `GenerateSessionTitleStatus::Skipped`. A user who clicked "generate title" therefore got a silent no-op, with nothing saying the session state could not be read. It now propagates, and the three `resolve_title_input_for_session` call sites — all already in `AppResult` functions — carry it.

`resolve_native_session`'s `locate_transcript_for` lookup gets the same treatment its `live_transcript_checked` sibling already had: a failed lookup is logged rather than folded into "no transcript".

## Why the branch was split

The original revision does not rebase: 34 conflict hunks across `agent_history.rs`, `agent_resume.rs`, `agent_resume_persister.rs`, `commands.rs`, `session_titles.rs`, and `todos.rs`, because `main` has independently reworked all of them. Each hunk is a judgement between two implementations rather than a merge, so it has to be re-authored — and 1,000 lines of that across the resume and history flows is not reviewable in one pass.

Remaining slices, in order:

1. ~~`session_titles.rs`~~ — this PR
2. `todos.rs::locate_transcript_for` and its callers
3. `agent_resume.rs` + `agent_resume_persister.rs`
4. `agent_history.rs` — the largest
5. the frontend half (`AgentResumeModal.tsx`, `App.tsx`, `RightPanel.tsx`) once the backend contract settles

## Test plan

- [x] `cargo test --manifest-path src-tauri/Cargo.toml -p acorn --lib` — 859 passed, 1 ignored
- [x] Negative control: restoring the swallow makes `chat_title_input_reports_an_unreadable_state_file` fail.

The test uses a `_from_dir` seam (`load_chat_session_state_from_dir` widened to `pub(crate)`) rather than `ACORN_DATA_DIR` so it does not race other env-mutating tests.

### DIFF
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -7973,7 +7973,7 @@ fn session_title_readiness_inner(
 ) -> AppResult<SessionTitleReadinessResult> {
     let id = parse_id(&id)?;
     let session = state.sessions.get(&id)?;
-    let title_input = resolve_title_input_for_session(&session, id);
+    let title_input = resolve_title_input_for_session(&session, id)?;
     let transcript_id = title_input
         .as_ref()
         .map(|input| input.transcript_id.as_str());
@@ -8042,7 +8042,7 @@ fn generate_session_title_inner(
 ) -> AppResult<GenerateSessionTitleResult> {
     let id = parse_id(&id)?;
     let session = state.sessions.get(&id)?;
-    let title_input = resolve_title_input_for_session(&session, id);
+    let title_input = resolve_title_input_for_session(&session, id)?;
     let transcript_id = title_input
         .as_ref()
         .map(|input| input.transcript_id.as_str());
@@ -8067,7 +8067,7 @@ fn generate_session_title_inner(
     let generated =
         crate::session_titles::generate_title(&ai, prompt.as_deref(), &title_input.title_context)?;
     let latest = state.sessions.get(&id)?;
-    let current_title_input = resolve_title_input_for_session(&latest, id);
+    let current_title_input = resolve_title_input_for_session(&latest, id)?;
     let current_transcript_id = current_title_input
         .as_ref()
         .map(|input| input.transcript_id.as_str());
@@ -8120,11 +8120,11 @@ fn can_store_generated_session_title(
 fn resolve_title_input_for_session(
     session: &Session,
     id: Uuid,
-) -> Option<crate::session_titles::ResolvedTitleInput> {
+) -> AppResult<Option<crate::session_titles::ResolvedTitleInput>> {
     if session.mode == SessionMode::Chat {
         crate::session_titles::resolve_chat_title_input(id)
     } else {
-        crate::session_titles::resolve_title_input(id)
+        Ok(crate::session_titles::resolve_title_input(id))
     }
 }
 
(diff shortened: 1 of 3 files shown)


im-ian:
Reviewed while landing the access-error series. Leaving this open — the concern is real and largely unaddressed on `main`, but the branch does not survive a rebase and the replacement is too big to write blind.

**Still real on `main`.** The swallows this PR targets are mostly still there. Two concrete ones:

```rust
// session_titles.rs
pub fn resolve_chat_title_input(session_id: uuid::Uuid) -> Option<ResolvedTitleInput> {
    let state = crate::persistence::load_chat_session_state(&session_id.to_string()).ok()?;
    ...
}

// session_titles.rs — resume transcript lookup
todos::locate_transcript_for(&session_id.to_string())
    .ok()
    .flatten()
    .map(|path| agent_resume::LiveTranscript { ... })
```

Both turn "could not read it" into "there is nothing here": a session whose state file is unreadable silently gets no generated title, and a transcript that cannot be located because of an access error looks like a session with no transcript to resume. Counting the same shape across the files this branch touches: `agent_history.rs` ~46, `todos.rs` ~15, `agent_resume.rs` ~10, `agent_resume_persister.rs` ~7.

**Why it does not rebase.** 34 conflict hunks across `agent_history.rs`, `agent_resume.rs`, `agent_resume_persister.rs`, `commands.rs`, `session_titles.rs`, and `todos.rs`. `main` has independently reworked all of those since this branch was cut, so most hunks are a per-site judgement between two different implementations rather than a mechanical merge — the same situation that made #801, #800, #810, and #818 need re-authoring rather than rebasing.

**Suggested split.** This is really several changes wearing one branch. Landing them separately would each be reviewable and testable on its own:

1. `session_titles.rs` — smallest and self-contained (the two sites above).
2. `todos.rs::locate_transcript_for` and its callers.
3. `agent_resume.rs` + `agent_resume_persister.rs`.
4. `agent_history.rs` — by far the largest, worth its own pass.
5. The frontend half (`AgentResumeModal.tsx` +152, `App.tsx`, `RightPanel.tsx`) once the backend contract settles.

Happy to take these one at a time, starting with (1), on your go-ahead. I did not want to rewrite 1,000 lines across the resume and history flows without you seeing the split first.

im-ian:
All remaining slices have landed. Final mapping, for anyone following this thread:

1. `session_titles.rs` — this PR.
2. `todos.rs::locate_transcript_in` (sibling-safe scan) plus the `RightPanel` Todos tab — #844.
3. `agent_resume.rs` + `agent_resume_persister.rs` — landed on `main` via #834, which introduced the `_checked` variants and routed the command paths through them.
4. `agent_history.rs` — also #834; every remaining swallowing wrapper there is now `#[cfg(test)]`-only.
5. Frontend (`AgentResumeModal.tsx`, `App.tsx`, locales, e2e) — #845.

PR #793, opened 2026-08-12
fix(shell): protect staged init files from linked paths

im-ian:
## Summary

This prevents Acorn staging its managed zsh initialization files through linked filesystem entries.

1. **Plain managed directory** — reject a `shell-init` path that resolves to a symlink or another non-directory entry before publishing any managed files.
2. **Atomic leaf replacement** — publish each staged zsh file through the existing same-directory atomic writer, replacing symlink or hardlink directory entries instead of writing through them.
3. **Regression coverage** — prove linked leaves cannot modify their peer files and a linked managed directory remains untouched.

## Expected impact

| Scenario | Before | After |
| --- | --- | --- |
| Managed leaf replaced by symlink/hardlink | Startup writes could modify the linked peer | Peer remains unchanged; the managed directory entry is replaced atomically |
| Managed `shell-init` directory is linked | Files could be populated outside the managed directory | Setup fails closed and zsh falls back to normal cwd tracking |
| Normal zsh startup | Four files rewritten directly | Same contents published atomically |

## Design notes

- This reuses `acorn_platform::fs::write_atomic`, which stages in the destination directory and renames over the directory entry.
- The existing caller already treats shell-init setup as optional and logs a warning before falling back to focus-based cwd refresh.

## Test plan

- [x] `cargo test --manifest-path src-tauri/Cargo.toml -p acorn shell_init::tests -- --nocapture` — 11 passed.
- [x] `cargo test --manifest-path src-tauri/Cargo.toml --workspace` — passed; the first run exposed one unrelated wrapper-process timing flake, whose targeted rerun and the full rerun both passed.
- [x] `cargo clippy --manifest-path src-tauri/Cargo.toml -p acorn --lib` — completed successfully.
- [x] `git diff --check` — clean.

## Notes

- The transient `agent_wrappers::tests::codex_wrapper_maps_current_tui_user_turn_to_turn_start` failure produced an empty watcher output; it passed immediately in isolation and in the complete rerun. The touched shell-init code is not on that wrapper test path.
- Clippy reports existing repository warnings unrelated to this PR.

### DIFF
--- a/src-tauri/src/shell_init.rs
+++ b/src-tauri/src/shell_init.rs
@@ -55,10 +55,20 @@ pub fn ensure_shell_init_dir() -> io::Result<PathBuf> {
 fn ensure_shell_init_dir_at(base: &Path) -> io::Result<PathBuf> {
     let dir = base.join(SHELL_INIT_DIR_NAME);
     fs::create_dir_all(&dir)?;
-    fs::write(dir.join(ZSHENV_NAME), ZSHENV_BODY)?;
-    fs::write(dir.join(ZPROFILE_NAME), ZPROFILE_BODY)?;
-    fs::write(dir.join(ZSHRC_NAME), ZSHRC_BODY)?;
-    fs::write(dir.join(ZLOGIN_NAME), ZLOGIN_BODY)?;
+    let metadata = fs::symlink_metadata(&dir)?;
+    if !metadata.file_type().is_dir() {
+        return Err(io::Error::new(
+            io::ErrorKind::InvalidData,
+            format!(
+                "shell init path is not a plain directory: {}",
+                dir.display()
+            ),
+        ));
+    }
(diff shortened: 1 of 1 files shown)


im-ian:

This PR guarded the staged zsh init files against symlinked/hard-linked leaves and a symlinked managed directory. `main` now does both, and does them more strictly:

- directory guard — `ensure_plain_directory(&dir)?` replaces the `symlink_metadata` + `is_dir` check proposed here.
- leaf replacement — `acorn_platform::fs::write_atomic_private(...)` replaces the `write_atomic(...)` proposed here, so the staged files are additionally created with private permissions.

`main` also already carries the matching regression tests: `shell_init_write_replaces_symlink_without_touching_target` and `shell_init_rejects_symlinked_managed_directory`.

Rebasing this branch would have replaced `write_atomic_private` with `write_atomic` and dropped the private-permission behaviour for no behavioural gain, so the branch is closed rather than landed.

PR #798, opened 2026-08-12
fix(settings): surface project settings access errors

im-ian:
## Summary

Project settings now distinguish a missing settings file from one that cannot be accessed, preventing updates from rebuilding state from an accidental empty map.

1. **Fail-closed loading:** read the settings file directly and treat only `NotFound` as an empty settings map.
2. **Error propagation:** preserve permission and other I/O failures for get, update, and remove call paths.
3. **Regression coverage:** verify an inaccessible parent directory returns `PermissionDenied` and leaves the existing settings file unchanged.

## Expected impact

| Area | Before | After |
| --- | --- | --- |
| Inaccessible settings file | Could appear missing through `Path::exists` | Returns the underlying access error |
| Subsequent settings updates | Could overwrite from an empty in-memory map | Stop before the atomic write |
| Missing settings file | Starts with an empty map | Unchanged |

## Design notes

The JSON schema, parse-error behavior, and atomic writer are unchanged. The loader now follows the same missing-versus-inaccessible contract expected by mutating callers without adding a separate state layer.

## Test plan

- [x] `rustfmt --edition 2021 --check src-tauri/src/project_settings.rs`
- [x] `cargo test --lib project_settings::tests` — 6 passed
- [x] `cargo test --workspace --quiet -- --test-threads=1` — passed
- [x] `cargo clippy --workspace --all-targets` — passed with existing warnings

## Notes

The workspace-wide `cargo fmt --all -- --check` remains blocked by a pre-existing formatting difference in `src-tauri/src/pty_env.rs`; this PR does not modify that file.

### DIFF
--- a/src-tauri/src/project_settings.rs
+++ b/src-tauri/src/project_settings.rs
@@ -73,10 +73,17 @@ fn settings_path() -> AppResult<PathBuf> {
 
 fn load_all() -> AppResult<SettingsMap> {
     let path = settings_path()?;
-    if !path.exists() {
-        return Ok(SettingsMap::new());
-    }
-    let bytes = fs::read(&path)?;
+    load_all_from(&path)
+}
+
+fn load_all_from(path: &Path) -> AppResult<SettingsMap> {
+    let bytes = match fs::read(path) {
+        Ok(bytes) => bytes,
+        Err(err) if err.kind() == std::io::ErrorKind::NotFound => {
+            return Ok(SettingsMap::new());
+        }
+        Err(err) => return Err(err.into()),
+    };
     serde_json::from_slice::<SettingsMap>(&bytes)
         .map_err(|err| AppError::Other(format!("failed to parse project settings: {err}")))
 }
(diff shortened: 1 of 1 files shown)


im-ian:

This PR replaced the `path.exists()` pre-check in `load_all` with an error-preserving read so a permission failure could no longer be reported as "no settings". `main` now does exactly that, through `persistence::read_bounded_regular_file`:

```rust
let bytes = match persistence::read_bounded_regular_file(&path, MAX_PROJECT_SETTINGS_FILE_BYTES) {
    Ok(bytes) => bytes,
    Err(AppError::Io(err)) if err.kind() == std::io::ErrorKind::NotFound => {
        return Ok(SettingsMap::new());
    }
    Err(err) => return Err(err),
};
```

A blocked parent directory therefore already surfaces `PermissionDenied` instead of an empty settings map — the exact case this branch's regression test covers.

`main`'s version is also stricter than the one proposed here: it bounds the read at `MAX_PROJECT_SETTINGS_FILE_BYTES` and rejects non-regular files, neither of which the plain `fs::read` in this branch does. Rebasing this branch would have traded those protections for no behavioural gain.

Only the `load_all_from(&Path)` seam and its test are left unique to this branch. If the permission case is worth pinning down in a test, it is better added against `main`'s `read_bounded_regular_file` path in a fresh PR than by landing the weaker read.

PR #807, opened 2026-08-12
fix(worktrees): preserve access errors in authorization + probes

im-ian:
## Summary

`list_worktree_paths` is authoritative: `authorize_registered_worktree` denies anything missing from it, and `authorize_project_session_cwd` decides scope from it. A registration Acorn could not read silently dropped out, so a worktree that really *is* registered reported "worktree is not registered for the project".

## What I got wrong first, and the correction

My first attempt was the obvious one — propagate libgit2's errors instead of `if let Ok(wt) = repo.find_worktree(name)`. **It does not fix the gap.** With one registration directory chmod'd to `0o000`, the listing still came back `Ok([])`: `git_worktree_list` drops an unreadable registration during *enumeration*, so the name never reaches the caller and there is no error to propagate.

That is exactly what this branch's original description said, and I should have taken it at face value:

> `git_worktree_list` can silently skip unreadable registration directories. The shared listing helper therefore inspects the common-dir registration entries first and resolves each one through libgit2 as a single fail-closed batch.

So the implementation now follows the original design: enumerate `<commondir>/worktrees` directly, resolve each entry through libgit2, and fail the batch if any step fails.

## Why failing the batch is safe

A **stale** registration — checkout deleted, registration intact — still resolves and yields its recorded path. `list_worktree_paths_still_reports_a_stale_registration` pins that, so `stage_remove_worktree_at_path` can still prune it. Only a genuinely unreadable registration fails the listing.

`authorize_project_session_cwd` also stops discarding every `canonicalize` failure: a missing checkout is skipped (it cannot contain the cwd), anything else is an error.

## What is not included

The rest of the original branch is already on `main` and stronger — `canonical_existing_path` in the authorization helpers, plus #782's `path_entry_exists` in `stage_remove_worktree_at_path`. `registered_repository_roots` keeps its tolerant `match` + `tracing::debug!`: it degrades to *fewer* scope roots, which is fail-closed, and hard-failing there would take out project listing for one bad repo.

## Test plan

- [x] `cargo test --manifest-path src-tauri/Cargo.toml -p acorn --lib` — 860 passed, 1 ignored
- [x] Negative control: with `list_worktree_paths` restored to `main`'s version, `list_worktree_paths_fails_on_an_unreadable_registration` fails.
- [x] The unreadable test blocks a single registration directory, not `.git/worktrees` itself — blocking the parent makes `repo.worktrees()` fail, which `main` already propagated, so the test would have passed vacuously.

Both new tests skip their assertions when the sandbox still permits access to a `0o000` directory.

### DIFF
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -661,12 +661,27 @@ pub(crate) fn authorize_project_session_cwd(repo: &Path, cwd: &Path) -> AppResul
     if path_is_inside(cwd, repo) {
         return Ok(());
     }
-    if worktree::list_worktree_paths(repo)?
-        .into_iter()
-        .filter_map(|path| path.canonicalize().ok())
-        .any(|worktree| path_is_inside(cwd, &worktree))
-    {
-        return Ok(());
+    for worktree_path in worktree::list_worktree_paths(repo)? {
+        // A worktree whose checkout is gone cannot contain `cwd`, so skipping it
+        // is correct. Any other failure is not evidence of absence, and letting
+        // it through would deny a cwd that really is inside a registered
+        // worktree.
+        let canonical = match worktree_path.canonicalize() {
+            Ok(path) => path,
+            Err(error) if error.kind() == std::io::ErrorKind::NotFound => continue,
+            Err(error) => {
+                return Err(AppError::Io(std::io::Error::new(
+                    error.kind(),
+                    format!(
+                        "failed to inspect registered worktree {}: {error}",
+                        worktree_path.display()
+                    ),
+                )))
+            }
+        };
+        if path_is_inside(cwd, &canonical) {
+            return Ok(());
+        }
     }
     Err(AppError::InvalidPath(format!(
         "cwd is outside the registered project and its worktrees: {}",
--- a/src-tauri/src/worktree.rs
+++ b/src-tauri/src/worktree.rs
@@ -375,14 +375,58 @@ fn configured_worktree_base_commit<'repo>(
 /// excluded; libgit2's `worktrees()` only reports `.git/worktrees/<name>`
 /// entries. That matches what we want — `claude -w` always adds a linked
 /// worktree, never modifies the main one.
+/// List every registered linked worktree path.
+///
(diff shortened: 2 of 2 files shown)


im-ian:
Reviewed while landing the access-error series. Leaving this open rather than rebasing it — the branch as written would undo hardening `main` has since gained, and the part that is still real is a policy call.

**Already on `main`, in a stronger form.** The authorization helpers all fail closed today: `authorize_registered_project_root`, `authorize_registered_repository`, and `authorize_registered_worktree` go through `canonical_existing_path` and propagate its error, and `authorize_registered_worktree` additionally rejects non-normalized or relative paths before touching the filesystem. `stage_remove_worktree_at_path` was the one place still deciding from `Path::exists()`, and #782 covers that.

**What is still real.** `list_worktree_paths` builds its result with two silent skips:

```rust
for name in names.iter().filter_map(|name| name.ok().flatten()) {
    if let Ok(wt) = repo.find_worktree(name) {
        paths.push(wt.path().to_path_buf());
    }
}
```

A registration whose name cannot be read, or that `find_worktree` cannot open, drops out of the list. Callers treat the list as authoritative, so that worktree then reports "worktree is not registered for the project" (`authorize_registered_worktree`) or "cwd is outside the registered project and its worktrees" (`authorize_project_session_cwd`) — a misleading denial for a worktree that is registered.

**Why I did not just change it.** The direction is fail-closed, so this is a usability problem rather than a security one, and the fix is not obviously "propagate the error": doing that makes a single broken registration fail the whole listing, which would take out the project view instead of one worktree. `main`'s skip may well be deliberate tolerance for damaged registrations.

**Options:**

- Keep the listing tolerant but return the skipped names alongside it, so the authorization helpers can say "this worktree's registration is unreadable" instead of "not registered".
- Propagate the error and accept that one bad registration fails the listing.
- Leave as is.

The first needs a decision on the return shape before it can be written. Happy to implement whichever you pick.

The branch's other half — reworking `list_worktree_paths` into `scan_registered_worktree_paths` plus `is_registered_worktree_root` / `is_inside_registered_worktree` — can be rebuilt on top of whichever option you choose; it does not survive a rebase as written.

PR #808, opened 2026-08-12
fix(git): preserve diff image access failures

im-ian:
## Summary

`read_workdir_image` collapsed every outcome to `None`, so an image Acorn could not read was indistinguishable from a binary file with no preview. The user saw an empty pane and no way to tell the two apart.

The outcomes are now split:

- **silent** (`Ok(None)`) — policy rejections (symlink, outside the workdir, bad relative path), the size budget, and a file that is simply gone;
- **reported** (`Err`) — a file that is present but unreadable, surfaced as `DiffFile.image_error` and rendered by the image pane in place of the "no preview" text.

## Why not fail the whole diff

The earlier revision threaded `AppResult` up into the diff assembly, so one unreadable image failed the entire diff request. That is the wrong trade here, and the reason is in the call path: the workdir read only runs when the delta has no oid — untracked or staged-but-unwritten files. The realistic failure is therefore a **file vanishing between the diff being computed and the preview being read** (a build, a `git clean`), not a permission problem. Turning that common race into a failed diff — or even into a visible error — would be noise. `workdir_image_read_stays_silent_when_the_file_is_gone` pins the quiet path.

The earlier revision also dropped `same_workdir_file`, the post-open `dev`/`ino` re-check that proves the bytes came from the inode that passed the symlink and size checks. That is kept.

## Wire compatibility

`image_error` is `#[serde(skip_serializing_if = "Option::is_none")]`, so nothing changes on the wire in the normal case. `unified_diff.rs` sets it to `None`.

## Test plan

- [x] `cargo test --manifest-path src-tauri/Cargo.toml -p acorn --lib` — 858 passed, 1 ignored
- [x] `pnpm run typecheck` — clean
- [x] `pnpm run test` — 1450 passed (123 files)
- [x] Negative control: restoring the silent arm on `File::open` makes `workdir_image_read_reports_an_unreadable_file` fail.

### DIFF
--- a/src-tauri/src/git_ops.rs
+++ b/src-tauri/src/git_ops.rs
@@ -119,6 +119,11 @@ pub struct DiffFile {
     pub old_image: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub new_image: Option<String>,
+    /// Why this file's preview is missing, when the reason was an access
+    /// failure rather than policy or the file simply being gone. Absent in the
+    /// normal case, so existing consumers are unaffected.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub image_error: Option<String>,
     pub is_image: bool,
 }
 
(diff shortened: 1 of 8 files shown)


im-ian:
Reviewed while landing the access-error series. Leaving this one open rather than rebasing it, because the residual change is a policy call rather than a defect fix.

**What is still real on `main`.** `read_workdir_image` collapses every failure to `None`, so an unreadable working-tree image is indistinguishable from "no image here". That much is unchanged.

**Why this revision is not obviously the right fix.**

1. *It fails the whole diff.* The branch threads `AppResult` all the way up through `image_data_uri_workdir_bounded` into the diff assembly (`let images: AppResult<(Option<String>, Option<String>)> = …` → `return Err(error)`). One image the user cannot read therefore takes down the entire diff view, where today the diff still renders and only that preview is missing. For a best-effort preview inside a larger payload that trade looks like a net loss.

2. *It drops a TOCTOU guard.* `main` re-checks `same_workdir_file(&metadata, &open_meta)` after opening, so the bytes read are provably from the inode that passed the symlink/size checks. This revision removes that check.

**Options, in the order I would rank them:**

- Keep the diff succeeding and mark the affected file in the payload so the UI can render "preview unavailable — access denied" instead of nothing. Costs a payload field; keeps the diff usable.
- Land the propagation as written, accepting that one unreadable image fails the diff — but re-add `same_workdir_file` first.
- Close this and leave the current best-effort behaviour.

Happy to implement whichever you prefer; option 1 needs a decision on the payload shape before it can be written.

PR #809, opened 2026-08-12
fix(todos): preserve transcript access failures in the panel

im-ian:
## Summary

This PR stops Claude todo transcript access failures from being presented as an empty todo list.

1. **Checked transcript discovery** — distinguishes a genuinely missing projects root, project directory, or transcript from canonicalization, traversal, metadata, and permission failures.
2. **Sibling-safe lookup** — remembers failures while scanning project directories so a later readable transcript can still win; if no match is provable, the first access failure is returned.
3. **Visible panel errors** — keeps the Todos tab available when its request fails and renders the existing request error instead of silently hiding the tab.
4. **Regression coverage** — covers missing roots, unreadable project directories, and the frontend error presentation.

## Expected impact

| Scenario | Before | After |
| --- | --- | --- |
| Claude projects/transcript path is unreadable | Todos silently become empty and the tab disappears | The Todos tab remains available and shows the access error |
| Projects root or requested transcript is absent | Empty todo list | Empty todo list (unchanged) |
| Another project directory errors but a valid transcript is found | Lookup could discard the failure as empty | The readable exact match still wins |
| Transcript safety and replay budgets | Enforced | Unchanged |

## Design notes

- `NotFound` remains an optional lookup result because directories and transcripts may legitimately be absent or disappear during polling.
- Non-`NotFound` entry errors are retained while the bounded scan continues. This avoids letting an unrelated unreadable sibling mask a valid later match, without claiming an empty result when the scan was incomplete.
- The frontend hook already stored request failures. The rendering path now treats that error as a reason to retain the otherwise data-driven Todos tab.
- Status polling continues to treat marker-less transcript discovery as a best-effort heuristic; this PR changes the explicit Todos request, whose `AppResult` and UI already have an error contract.

## Test plan

- [x] `cargo test --manifest-path src-tauri/Cargo.toml --lib todos::tests` — 17 passed.
- [x] `pnpm exec vitest run src/components/RightPanel.test.tsx` — 21 passed.
- [x] `pnpm typecheck`
- [x] `pnpm build:frontend`
- [x] `cargo test --manifest-path src-tauri/Cargo.toml --lib -- --skip top_level_wrappers_create_one_provider_independent_invocation_owner --skip codex_wrapper_maps_current_tui_user_turn_to_turn_start` — 748 passed, 1 ignored, 2 intentionally filtered.
- [x] `cargo test --manifest-path src-tauri/Cargo.toml --lib agent_wrappers::tests::codex_wrapper_maps_current_tui_user_turn_to_turn_start -- --exact --nocapture` — passed on immediate isolated rerun.
- [x] `cargo clippy --manifest-path src-tauri/Cargo.toml --lib --tests` — passed with pre-existing warnings only.
- [x] `rustfmt --edition 2021 --check src-tauri/src/todos.rs`
- [x] `git diff --check`

## Notes

The repository's known `top_level_wrappers_create_one_provider_independent_invocation_owner` local harness hang remains excluded. One unrelated wrapper shell-capture test returned an empty fixture once during the first full run and passed immediately in isolation; the clean full run filters that pre-existing flake.

### DIFF
--- a/src-tauri/src/todos.rs
+++ b/src-tauri/src/todos.rs
@@ -20,7 +20,7 @@
 use serde::{Deserialize, Serialize};
 use std::collections::{BTreeMap, HashMap};
 use std::fs::{File, Metadata};
-use std::io::{BufRead, BufReader, Read};
+use std::io::{self, BufRead, BufReader, Read};
 use std::path::{Path, PathBuf};
 use uuid::Uuid;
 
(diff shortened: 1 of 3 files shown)


im-ian:
Superseded by #791, which already owns the transcript-discovery contract. The non-overlapping RightPanel error presentation and regression test were moved to #791 in commit 62be68b.

PR #857, opened 2026-08-24
perf(terminal): coalesce TUI scroll IPC so Claude CLI stays responsive

im-ian:
## Why

Scrolling inside Claude CLI (and other alt-screen TUIs) froze Acorn. The wheel is app-owned, not xterm viewport scroll. Each tick was:

1. up to 8 mouse reports → 8 `pty_write` IPCs, often on the UI thread via a blocking daemon RPC
2. a full TUI redraw coming back as 4KB PTY frames, each becoming a WKWebView Channel `eval` (and a fetch round-trip above 1KB)
3. extra JS work on that firehose (CSI-stripping permission detector, 2 Hz whole-process `pty_cwd`)

## What

- Coalesce PTY **output** in the app: leading-edge send for keystroke echo, then 8ms / 64KB catch-up window. Daemon protocol unchanged.
- Make `pty_write` async + `spawn_blocking`, serialized per session so reports cannot reorder.
- Join stdin writes that land in the same turn into one `pty_write` (microtask coalescer).
- Skip the permission-string detector while a TUI owns the pointer; only inspect the last 1KB of a chunk otherwise.
- Probe live cwd every 5s instead of 500ms.

Left alone on purpose: DOM renderer (IME), `MAX_LINES_PER_EVENT` (native-equivalent reporting), per-write `term.refresh`.

## Test plan

- [x] `pnpm exec vitest run src/lib/ptyWriteCoalescer.test.ts src/lib/permissionWarmup.test.ts src/lib/terminalWheelScroll.test.ts`
- [x] `pnpm run typecheck`
- [x] `cargo test --lib pty_output`
- [x] `cargo test --lib pty_write_does_not_infer`
- [ ] Scroll a long Claude Code conversation in Acorn: app chrome stays live, wheel still moves about one report per line
- [ ] Type in a normal shell: echo still feels immediate (leading-edge output flush)

### DIFF
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -8777,7 +8777,7 @@ fn decode_base64(input: &str) -> Option<Vec<u8>> {
 }
 
 #[tauri::command]
-pub async fn pty_spawn<R: Runtime>(
+pub async fn pty_spawn<R: Runtime + 'static>(
     app: AppHandle<R>,
     state: State<'_, AppState>,
     session_id: String,
@@ -8813,7 +8813,7 @@ pub async fn pty_spawn<R: Runtime>(
 }
 
 #[allow(clippy::too_many_arguments)]
-fn pty_spawn_blocking<R: Runtime>(
+fn pty_spawn_blocking<R: Runtime + 'static>(
     app: AppHandle<R>,
     state: AppState,
     session_id: String,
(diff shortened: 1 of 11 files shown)


im-ian:
Review (local) flagged two correctness bugs; both are in 86120701:

1. `pty_write` cloned `AppState`, which copied the DashMap, so the per-session lock never actually serialized writes. Locks are now `Arc<DashMap<…, Arc<tokio::sync::Mutex<()>>>>` and acquired with `lock().await` before `spawn_blocking`.
2. `unsubscribe` flushed the hold buffer into the outgoing Channel and then removed without a token predicate, so a remount could drop up to 64KB of TUI output. Unsubscribe is token-matched only; the 8ms timer delivers to the live subscriber.

IME E2E assertions now count syllables in the joined write stream, since same-turn `pty_write`s coalesce.

PR #885, opened 2026-09-07
fix(sessions): keep Grok working when monitors skip the manifest

im-ian:
## Summary

Grok writes `turn_completed` as soon as its TUI returns to the prompt, even while a live monitor is still watching CI. #877 kept those sessions **Working** by reading `background_tasks_manifest.json`. Grok does not always write that sibling file, so Acorn still showed **need input** in the middle of unfinished monitor work.

Keep the session **Working** when `updates.jsonl` has a `task_backgrounded` monitor (`kind: monitor`, `monitor_description`, `[monitor]` display command, or a `monitor-call-*` output file) that has not yet `task_completed`. Walk the full transcript, not just the 256 KiB status tail, because a long tool payload can push the original `task_backgrounded` line out of the window.

Fire-and-forget shells (`npm run dev`, leftover `vite`) still map to WaitingForInput.

## Changes

- Recognize Grok monitors from transcript `task_backgrounded` events, not only `background_tasks_manifest.json`.
- Scan `updates.jsonl` for unmatched monitor tasks so a spawn line that aged out of the 256 KiB tail still keeps Working.
- Leave unmatched bash `task_backgrounded` events as WaitingForInput after `turn_completed`.

## Test plan

- [x] `cargo test -p acorn-transcript -p acorn-session --lib` (236 passed)
- [x] Repro case: `task_backgrounded` + `monitor_description` + `turn_completed`, no manifest → Working
- [x] Same monitor after 300 KiB of later lines (outside the status tail) → Working
- [x] Monitor that later `task_completed` → WaitingForInput
- [ ] CI (Rust format, Frontend, E2E shards)

## Changelog category

`fix`

### DIFF
--- a/src-tauri/crates/acorn-session/src/status.rs
+++ b/src-tauri/crates/acorn-session/src/status.rs
@@ -30,7 +30,9 @@
 //!   WaitingForInput as an interrupted turn.
 //! - last `sessionUpdate=turn_completed` -> WaitingForInput while the Grok
 //!   process remains live (the TUI has returned to its prompt), unless a live
-//!   monitor or a running child agent is still in flight.
+//!   monitor or a running child agent is still in flight. Monitors come from
+//!   `background_tasks_manifest.json` or unmatched `task_backgrounded` events
+//!   in `updates.jsonl` when Grok omitted the sibling manifest.
 //! - user/agent/thought chunks, tool call updates, `task_backgrounded`, and
 //!   `subagent_spawned` -> Working.
 //! - `task_completed` / `subagent_finished` with `will_wake=true` -> Working.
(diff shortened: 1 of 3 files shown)


im-ian:
## Claude review (pre-merge)

Ran Claude Code (`opus`, high effort) against this PR. Verdict: **Approve with nits**.

Local `cargo test -p acorn-transcript -p acorn-session`: 236 pass. Heuristic also checked against 494 real `task_backgrounded`/`task_completed` events in `~/.grok/sessions` (48 monitors all have `monitor_description` + `monitor-call-*`; 446 bash tasks have neither).

### Bug — unmatched monitors pin Working forever

`updates.jsonl` is append-only. If a monitor dies without `task_completed` (quit / reboot / SIGKILL), the orphaned `task_backgrounded` never clears. Unlike the sibling manifest, you cannot delete your way out.

Confirmed on this machine: session `01a0461d-…` is 9.9 days old with an unmatched monitor ("Wait for PR 870 CI then squash-merge") and ends in `turn_completed`. Resume it and this PR would show Working forever — the inverse of the original symptom.

Suggested bound: ignore an unmatched monitor whose `output_file` is missing or whose mtime is older than the existing 24h `RECENCY_WINDOW_SECS`.

### Suggestions

1. Doc says the transcript scan is a fallback when the manifest is missing, but the code is an unconditional `||`.
2. Full-file scan has no byte cap and runs before the cheap 256 KiB tail check. Reorder: tail first, full scan only when `!tail.read_full`.
3. `read_line` into `String` aborts the rest of the file on invalid UTF-8; `read_tail` uses `from_utf8_lossy`.

### Nits

- No negative test using the real bash `task_backgrounded` shape (`call-<uuid>-N.log`).
- Pre-existing: CI does not run `-p acorn-transcript -p acorn-session`.

Not merging until we decide whether to land the orphan-mtime gate first.

im-ian:
Addressed the Claude review bug: unmatched `task_backgrounded` monitors now only keep Working when `output_file` exists and its mtime is within 24h. Missing or stale logs fall back to WaitingForInput.

Also reordered the followup checks so the cheap tail/subagent paths run before the full-file scan, and transcript lines are decoded with `from_utf8_lossy` like `read_tail`.

PR #889, opened 2026-09-09
fix(terminal): hide the overlay scrollbar that does not scroll

im-ian:
## Summary
- Hide xterm.js v6's overlay scrollbar. Wheel/trackpad scrollback and the existing scroll-to-bottom control still work; the thumb was a 6px auto-hide widget that intercepted clicks instead of driving the buffer.
- Also hide leftover native `overflow-y: scroll` on `.xterm-viewport` (xterm.css still sets it on an unused node).
- Drop unused `scrollbarSlider*` theme tokens now that the thumb is gone.
- Pin `js-yaml` to 4.3.2 via the existing pnpm override so Frontend CI `pnpm audit` can pass (GHSA-2883-xcg3-v3hh).

## Test plan
- [ ] Terminal with enough scrollback: no vertical thumb on the right edge.
- [ ] Trackpad/wheel still scrolls the buffer.
- [ ] Scrolled-away-from-bottom still shows the scroll-to-bottom button, and it still jumps to the live prompt.
- [ ] Sidebar / right-panel lists still hide *their* scrollbars (`acorn-no-scrollbar` is unchanged).
- [ ] `pnpm audit --audit-level=low` reports no known vulnerabilities.


im-ian:
Frontend CI failed on `pnpm audit` (GHSA-2883-xcg3-v3hh, js-yaml 4.3.1). Pushed a pin to 4.3.2 so the required Frontend job can finish; scrollbar change is unchanged.

PR #897, opened 2026-09-11
ci: run NSIS installer builds only on release

im-ian:
## Summary

Stop compiling the Windows NSIS installer and updater artifacts on every pull request. PR CI now type-checks the Rust workspace and keeps the cheap Windows-native tests; the release workflow still produces NSIS / DMG / updater bundles.

## Expected impact

Recent PR CI was gated by **Windows x64 NSIS smoke** (~12 min wall clock, of which ~8 min was `tauri build --bundles nsis`). After this change that job no longer packages an installer, so PR wall clock should fall back to the E2E shards (~8 min) and we stop paying Windows minutes for a path `release.yml` already runs.

| Surface | PR CI now | Release |
| --- | --- | --- |
| `tsc` + Vitest + Vite production bundle | yes (Vite is a few seconds; it still catches CSS/assets `tsc` misses) | `pnpm run build` |
| `cargo check --workspace --all-targets` | Linux + Windows | full link/bundle |
| Host-side `cargo test` | 8 crates on Linux (was `acorn-transcript` only) | — |
| Windows native tests + debug `acornd` detach smoke | yes | — |
| NSIS / DMG / updater artifacts | no | `release.yml` |

## Design notes

- Dropping the Windows job entirely would have *weakened* typechecking: the main `acorn` crate was effectively compiled on PRs only as part of the NSIS `tauri build`.
- `cargo check` of the Tauri crate needs `externalBin` paths to exist (`tauri-build` fails closed otherwise). Linux CI writes empty sidecar placeholders; Windows CI stages real debug sidecars because it also smokes `acornd serve --detach`.
- Linux `cargo check` installs the usual Tauri webkit/gtk headers. That is compile-only; Acorn still does not ship a Linux installer.
- The debug `acornd` smoke covers the Windows detach lifecycle without a release-ci link.

## Follow-up (not in this PR)

- Gate Playwright `tests/e2e` on `tsc` (~70 existing errors).
- `noUncheckedIndexedAccess` on `src` (hundreds of errors).
- `clippy -D warnings` on leaf crates (at least `chunks_exact` in `acorn-updater-verify`).

## Test plan

- [x] `cargo check --manifest-path src-tauri/Cargo.toml --locked --workspace --all-targets` (macOS, with sidecar placeholders)
- [x] `cargo test --manifest-path src-tauri/Cargo.toml --locked -p acorn-agent -p acorn-paths -p acorn-platform -p acorn-transcript -p acorn-local-ipc -p acorn-ipc -p acorn-daemon -p acorn-updater-verify` (277 passed)
- [ ] GitHub Actions on this PR: Frontend, E2E, Windows native, Rust tests, Rust format

## Changelog category

`ci`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -203,12 +203,6 @@ jobs:
             throw "Windows debug sidecar build failed"
           }
 
-      # NSIS / updater packaging lives in the release workflow. Sidecars
-      # above satisfy tauri-build's externalBin existence check; this
-      # type-checks the Tauri crate and every `cfg(windows)` test module.
-      - name: Type check Rust workspace
-        run: cargo check --manifest-path src-tauri/Cargo.toml --locked --workspace --all-targets
-
       - name: Smoke detached acornd lifecycle
         shell: pwsh
         run: |
@@ -360,6 +354,14 @@ jobs:
           # code at 2 ("not running"). Do not let pwsh report that expected
           # status as the workflow step's exit code.
           $global:LASTEXITCODE = 0
+
+      # NSIS / updater packaging lives in the release workflow. Sidecars
+      # above satisfy tauri-build's externalBin existence check; this
+      # type-checks the Tauri crate and every `cfg(windows)` test module.
+      # Run after the detach smoke so a check failure still leaves the
+      # Windows lifecycle signal in the same job.
+      - name: Type check Rust workspace
+        run: cargo check --manifest-path src-tauri/Cargo.toml --locked --workspace --all-targets
 
   rust-tests:
     name: Rust tests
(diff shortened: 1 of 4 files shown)


im-ian:
Opus review of the first commit (`59d5face`): **CHANGES_REQUESTED**, blocking.

**Blocking:** `cargo check --workspace --all-targets` compiles `acorn` test targets on Windows for the first time. `antigravity_subagent_stop_emits_no_transition` called `notify_event_for_stdin`, which is `#[cfg(unix)]`. That was the only E0425 in the failed Windows job.

**Addressed in `91198369`:**
- Gate that test with `#[cfg(unix)]`, matching its neighbors.
- Run the debug `acornd` detach smoke *before* the workspace check so a type error cannot skip the lifecycle signal.

**Non-blocking (not in this PR):** NSIS / `release-ci` / `stage-windows` now first run at tag time. A scheduled or `workflow_dispatch` copy of the old throwaway-key NSIS smoke would cover that without putting it back on the PR critical path.

PR #914, opened 2026-09-16
fix(windows): give every agent a legacy-drive PWD

im-ian:
## Summary
Follow-up to #910/#912. Claude slug encoding was fixed, but the **process cwd** agents actually inherit was still `\\?\W:\repo` from `std::fs::canonicalize`. Codex then compared `payload.cwd` with exact `PathBuf` equality and silently failed to pair; Grok percent-encoded the verbatim string into the wrong session bucket; PowerShell showed `Microsoft.PowerShell.Core.FileSystem::\\?\W:\...`.

Strip the verbatim disk prefix at every agent cwd boundary:

- in-process and daemon PTY spawn
- Loop/Graph CLI `current_dir`
- live-process cwd used for Claude/Codex/Grok/Antigravity pairing
- Codex `payload.cwd` comparison and Grok cwd-bucket encoding

## Expected impact
- Shell PWD and agent process cwd are `W:\repo` instead of `\\?\W:\repo`.
- Codex resume/history pairing works when the JSONL stored `W:\repo` and the process cwd was verbatim.
- Grok session buckets match the CLI's encoding of a legacy drive path.

## Test plan
- [x] `cargo test -p acorn-paths --offline same_cwd`
- [x] `cargo test -p acorn-transcript --offline agent_process_cwd_strips`
- [x] `cargo test -p acorn-transcript --offline grok_cwd_bucket_percent`
- [x] `cargo test -p acorn-transcript --offline antigravity_owner_cursor_matches`
- [x] `cargo test -p acorn-transcript --offline find_recent_codex_jsonl_pairs`
- [x] `cargo fmt --all --check`
- [ ] CI green on this PR


jtf-ian:
Pushed rustls 0.23.45 so `cargo audit` clears RUSTSEC-2026-0285 (TLS 1.3 handshake). Unrelated to the cwd change; the security job only runs because this PR touches `Cargo.lock`.

PR #915, opened 2026-09-16
fix(windows): report Codex turn completion through a PowerShell notify shim

im-ian:
Codex agent status was never reported on Windows, and the reason was structural rather than a bug in the status pipeline.

`agent_wrappers` writes extension-less `/bin/sh` shims into `<data_dir>/agent-wrappers` and `pty_spawn` prepends that directory to `PATH` under `#[cfg(unix)]`. Windows resolves commands through `PATHEXT`, so such a file is never launched — the gate is correct, but nothing took its place. Meanwhile `inject_agent_hook_env` is not gated, so every Windows PTY still received `ACORN_AGENT_HOOK_URL` / `ACORN_AGENT_HOOK_TOKEN` / `ACORN_AGENT_HOOK_SESSION_ID`: a hook channel that advertised itself and had no consumer. The failure warning sat inside the same `cfg(unix)` block, so nothing surfaced in the log either.

## What this adds

A PowerShell session function shadows `codex` instead of a PATH shim — no PATH manipulation, and no extra process between the console and Codex's TUI that would intercept Ctrl+C. It registers Codex's legacy `notify` channel, which reports turn completion.

Native hooks are deliberately out of scope. Codex requires every command hook to carry a `trusted_hash` fingerprint over its exact command string; that string necessarily differs on Windows and the fingerprint is derived inside Codex, so it cannot be precomputed here. **Approval-waiting therefore stays invisible on Windows.** There is no transcript-side fallback for it either — a Codex rollout JSONL carries only `user_message` / `task_started` / `task_complete` / `agent_message`, with no approval or tool-call events.

WSL is a separate and worse case that this does not fix: the Windows process scan only sees `wsl.exe` because the real binary runs inside the VM, and the transcript scan reads the Windows profile while the rollout lands in the distro's Linux home. Both status sources are blind, so the session reports a plain shell forever. Rather than let that read as Acorn silently failing, `isWslBridgedSession` names it in the status tooltip. It reuses `active_processes`, which the status poll already ships, so no new backend field is involved, and a real `status_reason` outranks the hint since its presence proves resolution worked.

## Three things worth reviewing closely

**No lifecycle id is sent, and that is load-bearing.** The first draft sent `X-Acorn-Codex-Lifecycle-Id` to match the POSIX contract, which made the whole feature a no-op: a lifecycle id routes the event into the Codex lane reducer, where `finish_turn(_, trusted_native: false)` drops any completion with no open turn, and nothing on Windows can open one (`native_prompt` needs native hooks; `jsonl_user` / `jsonl_task` come from the POSIX JSONL watcher). Every POST parsed cleanly and was then discarded. Without the header the event takes the unsequenced path, which exists for exactly this case — a wrapper that can report a boundary but cannot sequence it. Sequencing buys nothing here anyway: the shim runs Codex in the foreground, so a session has at most one shim-launched Codex at a time, and a nested Codex never receives the `-c notify` config. `windows_codex_notify_completion_ends_the_turn` and `windows_codex_notify_must_not_send_a_lifecycle_id` pin both halves.

**The shim is dot-sourced as script text, never through `-File`.** Execution policy gates script *files* and the Windows client default is `Restricted`, so `-File` would fail on a stock machine and put a red error at the top of every session. Launching the session with `-ExecutionPolicy Bypass` would have fixed it too, but that lowers the policy for everything the user subsequently runs in that shell rather than for one Acorn-owned file.

**The `notify` config is built from TOML literal strings.** Single quotes mean no escape processing, so Windows paths pass through verbatim and no double quote ever reaches the native argument — which is precisely what Windows PowerShell 5.1 mis-quotes. A path containing an apostrophe cannot be expressed that way, so the shim falls back to an unhooked run instead of emitting broken TOML.

## Verification

Rust and frontend: `cargo check --workspace --all-targets`, `cargo fmt --check`, `cargo test -p acorn --lib agent_wrappers` (60), the two new `agent_hooks` wire-format tests, `pnpm typecheck`, `pnpm test` (1614). CI is green on all six checks including `Windows native`.

The PowerShell was executed, against PowerShell 7.6.6 on macOS:

- **Both scripts parse.** `[Parser]::ParseFile` reports zero errors for each.
- **The notify receiver posts the right thing, once.** Driven against a local HTTP listener, the owned `agent-turn-complete` produced exactly one POST carrying `x-acorn-agent-hook-provider: codex`, `x-acorn-agent-hook-source: legacy_completion`, the session id, the token from the endpoint file, `x-acorn-codex-native-hooks-enabled: 0`, `content-type: application/json; charset=utf-8`, **no lifecycle header**, and the payload verbatim. These are byte-for-byte the headers `windows_codex_notify_completion_ends_the_turn` proves yield `Applied(WaitingForInput)`, so the chain is verified end to end.
- **Every rejection path holds.** A foreign `thread-id`, a non-completion `type`, a missing `thread-id`, and malformed JSON each produced zero POSTs and exit 0.
- **The shim builds the argv it should.** With a stub binary on PATH, `codex resume --last` reached the real command as `-c` / `notify=['<pwsh>','-NoProfile','-ExecutionPolicy','Bypass','-File','<notify.ps1>']` / `resume` / `--last` — one argument for the config, user arguments intact and in order. That generated string was then fed to a real `codex-cli` 0.148.0, which accepted it.
- **Both unhooked fallbacks fire.** An apostrophe in the wrapper path and an empty `ACORN_AGENT_HOOK_SESSION_ID` each ran the binary plainly with no `-c` and the user's arguments untouched.
- **A profile-defined `codex` survives.** With `function codex { … }` already defined, dot-sourcing the shim left it in charge and the user's function ran. With no `codex` on PATH at all, the shim declines to define anything, so the shell's own command-not-found is preserved.

Not covered: Windows itself. These ran on macOS PowerShell, so nothing here exercises `PATHEXT` resolution, execution policy, ConPTY signal handling, or Windows PowerShell 5.1's argument quoting — 5.1 is the reason the config avoids double quotes entirely. Worth a smoke test on a real Windows machine: run `codex` in an Acorn PowerShell session and confirm the tab leaves "working" at turn end without waiting for the 2-second transcript poll.


im-ian:
Follow-up: the "Not covered: Windows itself" caveat in the description no longer holds. The shim has now been exercised on a real Windows 11 Pro host running **Windows PowerShell 5.1.26100.9444** — the version whose native-argument quoting this design works around, and the host Acorn actually spawns there (no `pwsh` 7 on the box, so `shell_runtime` falls through to `powershell.exe`).

What the machine confirmed about the premise: `PATHEXT` is `.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC;.CPL`. No extension-less entry, and no `.PS1` either — so the POSIX wrapper-dir shim could never have been launched there regardless of `PATH`. The `#[cfg(unix)]` gate was load-bearing, not an oversight.

Results:

- **The config survives 5.1 *and* `cmd.exe`.** With a `.cmd` stub standing in for an npm-installed `codex` — the harshest path, since PowerShell quotes for the native call and then `cmd.exe` re-parses — `codex resume --last` reached the binary as `-c` / `notify=['C:/WINDOWS/System32/WindowsPowerShell/v1.0/powershell.exe','-NoProfile','-ExecutionPolicy','Bypass','-File','…/acorn-codex-notify.ps1']` / `resume` / `--last`. Single quotes, forward slashes and user argument order all intact. This was the riskiest decision in the PR and it holds on the platform it was made for.
- **The shim installs itself.** Dot-sourced into a 5.1 session, `function global:codex` is defined and takes over.
- **The receiver posts the right thing, once.** Driven the way Codex spawns it (a correctly quoted command line, as Rust's `Command` produces) and pointed at a listener over Tailscale, the owned `agent-turn-complete` produced exactly one POST: provider `codex`, source `legacy_completion`, the session id, the token read from the endpoint file, `x-acorn-codex-native-hooks-enabled: 0`, `content-type: application/json; charset=utf-8`, **no lifecycle header**, payload verbatim. Byte for byte the request `windows_codex_notify_completion_ends_the_turn` proves yields `Applied(WaitingForInput)`.
- **Every rejection path holds on 5.1 too.** Foreign `thread-id`, non-completion `type`, missing `thread-id` and malformed JSON each produced zero POSTs and exit 0.
- **Execution policy.** This particular host resolves to `Bypass`, so `-File` would have worked here. The dot-source-as-text path behaves identically on it, so the change is free where policy is permissive and is what saves a stock `Restricted` machine from an error at the top of every session.

One trap worth recording for anyone testing this by hand: invoking the receiver *from* PowerShell 5.1 (`& powershell.exe -File notify.ps1 $payload`) silently strips every double quote from the JSON, so the script sees `{type:agent-turn-complete,…}`, fails to parse, and exits 0 with no POST. That is the 5.1 quoting bug hitting the test harness rather than the shipped code — Codex spawns with a proper argv, which arrives intact. A first run of these tests showed zero POSTs for exactly this reason.

Still not covered: the full app path — Acorn spawning the PTY, and a real `codex` binary firing `notify` on its own (none is installed on that host). Signal handling through ConPTY is likewise untested.

PR #919, opened 2026-09-16
fix(windows): install the Codex shim over npm's generated launcher

im-ian:
Follow-up to #915, found by installing Codex on a real Windows host rather than reasoning about it.

The shim in #915 only shadowed `codex` when the name resolved to an `Application`. That guard came from review feedback — leave a user's own `codex` wrapper (a WSL bridge, a custom launcher) in charge rather than break a setup that works outside Acorn. The intent was right; the test was too broad.

`npm install -g @openai/codex` lays down three files side by side:

```
ExternalScript   C:\Users\…\AppData\Roaming\npm\codex.ps1
Application      C:\Users\…\AppData\Roaming\npm\codex.cmd
Application      C:\Users\…\AppData\Roaming\npm\codex
```

PowerShell resolves the `.ps1` first, so `Get-Command codex` returns an **ExternalScript** and the guard failed. On the most common Windows install of Codex the shim never installed at all — the whole feature was inert, in exactly the configuration most users have. (`npm` itself resolves the same way on that host: `npm.ps1`, ExternalScript.)

npm's generated launcher is not a user-authored wrapper; it is simply how npm installs a CLI on Windows. So refuse `Function` and `Alias` — which is the case the review actually raised — and shadow everything else.

## Verified on the Windows host

Windows 11 Pro, Windows PowerShell 5.1.26100.9444, `codex-cli` **0.154.0** installed via `npm install -g @openai/codex`:

- **Before the fix**: `Get-Command codex` → `ExternalScript`, guard fails, no shim.
- **After**: shim installs. `codex features list` runs *through* the shim and Codex accepts the injected `-c notify=[…]` config — exit 0, 140 feature rows. `codex --version` through the shim reports `codex-cli 0.154.0`.
- **User wrapper still wins**: with `function global:codex { … }` defined before the shim is dot-sourced, `codex resume --last` runs the user's function, unchanged.

## Also hardened here

npm installs a bare `codex` beside `codex.cmd`. Both report as `Application` and the bare one is a `#!/bin/sh` script Windows cannot execute, so `Select-Object -First 1` over that list leaned on PowerShell ordering `.cmd` first. It does — measured on the host, the Application list is `[codex.cmd, codex]` — but since this PR is what enables the shim on npm installs, that ordering now decides whether `codex` works at all in an Acorn session. `Get-AcornCodexBinary` filters to matches that have an extension, and both the guard and the per-call resolution go through it.

One limit is stated in the shim rather than left implied: a user-authored `codex.ps1` earlier on PATH is indistinguishable from npm's and is shadowed. A `.cmd`/`.exe` wrapper is unaffected — it resolves as the first Application, so the shim calls it and the user's launcher still runs.

A review point about `Get-Command` returning an array and making the Function/Alias exclusion vacuous did not reproduce: without `-All` it returns one command (`count=1`), and with a user function defined it returns a single `Function`, so the guard correctly evaluates to `False`.

## Not covered

That host has no Codex credentials (`~/.codex/auth.json` absent, no `OPENAI_API_KEY`), so no turn can complete and Codex never fires `notify` on its own. The receiver itself is already verified on that machine — #915's follow-up comment records a real POST from Windows PowerShell 5.1 carrying the exact headers `windows_codex_notify_completion_ends_the_turn` proves yield `Applied(WaitingForInput)` — but the final hop, Codex deciding to call it, still rests on the config being accepted rather than on an observed callback.


im-ian:
Review pass on this branch raised four points. Two were settled on the Windows host rather than by reading, since that is what produced the original bug.

**Resolving to npm's extension-less `codex` — hardened (`0698d3f0`).** npm installs a bare `codex` beside `codex.cmd`; both report as `Application` and the bare one is a `#!/bin/sh` script Windows cannot execute. The shim took `Select-Object -First 1` from that list, so it leaned on PowerShell ordering `.cmd` first. Measured on the host: `Application matches count: 2` → `[codex.cmd, codex]`, and `Select-Object -First 1` picks `codex.cmd`, so there is no live failure. But this PR is what enables the shim on npm installs in the first place, and the consequence if the order ever differs is `codex` broken in every Acorn PowerShell session. `Get-AcornCodexBinary` now filters to matches that have an extension, and both the guard and the per-call resolution go through it.

**Array-valued `Get-Command` making the Function/Alias exclusion vacuous — not reproduced.** The premise is that `Get-Command -Name codex` returns all three npm files even without `-All`. Measured on the host:

```
without -All : count=1 types=[ExternalScript]
with -All    : count=3 types=[ExternalScript, Application, Application]
```

It is `-All` that returns three. With a user function also defined, `Get-Command -Name codex` returns a single `Function`, and the guard evaluates to `False` — the user's function wins. Confirmed behaviourally as well: with `function global:codex { … }` defined first, `codex resume --last` runs the user's function.

**User-authored wrapper that is a script file — real, now documented.** A user's own `codex.ps1` earlier on PATH is indistinguishable from npm's and does get shadowed. A `.cmd`/`.exe` wrapper is *not* affected, though: it resolves as the first Application, so the shim calls it and the user's launcher still runs. Restricting the shadow to the npm prefix would close the `.ps1` case; it is not worth the extra failure mode here, so the limit is stated in the shim comment instead of left implied.

**Stale test comment — fixed.** It described the invariant this PR removes.

Re-verified on the host after the change: `Get-AcornCodexBinary` returns `…\npm\codex.cmd`, `codex features list` runs through the shim (exit 0, 140 rows), `codex --version` reports 0.154.0, and the profile-function case still wins.

Unrelated note for anyone running the suite locally: `cargo test -p acorn --lib agent_wrappers` flakes about one run in three or four on `codex_parent_guard_does_not_spawn_polling_sleep_processes` / `watcher_processes_exit_when_their_wrapper_is_terminated_directly`. Those spawn real `tail`/`pkill` processes and are load-sensitive. Checked against `origin/main` at the same parallelism — same rate, same tests — so it is pre-existing and not from this branch.

PR #932, opened 2026-09-17
fix(windows): stop verbatim paths from breaking session create/remove

im-ian:
## Summary
Windows에서 세션 생성·제거가 깨지던 문제를 고칩니다.

1. **verbatim 경로 (`\\?\`)** — canonicalize가 `\\?\D:\...`를 그대로 저장해서 프로젝트/세션 경로 비교가 어긋나고 생성이 막혔습니다. 저장·비교 모두 legacy drive path로 맞춥니다.
2. **remove `session not found`** — 낙관적 remove 후 `sessions` 의존 `useEffect`가 두 번째 `remove_session`을 다시 호출해 toast가 났습니다. in-flight 가드를 넣고 backend remove를 idempotent하게 만들었습니다.

## Test plan
- [x] `cargo test -p acorn-paths`
- [x] `pnpm typecheck`
- [x] `vitest run src/lib/pathUtils.test.ts --environment node`
- [ ] Manual Windows: drive path(예: `D:\…`)에서 세션 생성 + 제거

## Notes
- Release workflow는 의도적으로 꺼 둔 상태를 유지합니다. Windows E2E 확인 전 릴리즈하지 마세요.


im-ian:
Pushed `66ca1e1d` here — `cargo fmt` only. `Rust format` was the one red check; `Windows native` had failed earlier on the known flaky `powershell_conpty_accepts_input_resize_and_captures_output` ConPTY test and went green on a rerun, in a crate this branch does not touch.

The formatting drift was entirely inside files this branch already changes: an import reflow in `commands.rs` (`path_is_inside` into alphabetical position, lines rewrapped) and a stray blank line in `session.rs`. Identifier sets before and after are identical — no semantic change.

Touched this while the branch had been idle for four days with a one-command failure standing between it and a release that is blocked on it. Reverting the commit is harmless if you would rather redo it yourself.

im-ian:
Reviewed before merging, since I pushed the rustfmt commit here rather than authoring the change.

Independently corroborated on hardware: on Windows 11 with v1.37.1 installed from the release NSIS, the Agents panel showed `invalid path: project is not registered: \\?\C:\Users\jtf-dev` — the exact symptom this branch describes. That is what convinced me the release hold was justified, after I had wrongly called it unsubstantiated on the strength of GUI create/remove passing (my test box used `C:` and registered its project through the picker, so it never hit the mixed-form path).

What I checked:

- `owns_root` / `ProjectStore::{ensure,remove,rename,set_source_paths}` now resolve through `acorn_paths::same_cwd` instead of exact `PathBuf` key equality. The `iter().find(...)?.key().clone()` temporaries drop at the end of their statement, so the later `remove` / `get_mut` cannot deadlock against a held DashMap shard guard. The lookups go from hash to linear, which is irrelevant at project-count scale.
- `removeInFlightIdRef` in `App.tsx` mirrors the existing archive guard and clears on `!pendingRemove`, so a re-fired effect cannot issue the second invoke that produced the `session not found` toast.
- Replacing the local `normalizeWorkspacePath` with the shared `pathsEqual` drops a duplicated helper — net deletion in `Sidebar.tsx`.

One note, not a blocker: `ProjectStore::remove` now matches any root the project owns, `source_paths` included, where before it matched only the primary key. All four callers (`commands.rs:7001,7768,7894`, `lib.rs:85`) pass a root of the project they intend to remove, so this reads as a widening that is correct rather than risky — worth knowing it changed.

CI is 6/6 green including `Windows native`. Merging.

Not verified end to end in the installed app on Windows — that needs a build carrying this change, which is the remaining step before the release hold can come off.

PR #935, opened 2026-09-21
fix(windows): detect shell exit so terminals can restart

im-ian:
A daemon-backed terminal never learned its shell had exited on Windows. No `[process exited — press Enter to restart]` banner appeared, Enter did not respawn the shell, and the sidebar kept reporting the session as `Ready` on top of a dead PTY. The tab looked alive and was unusable.

## Root cause

`pty:exit:<id>` reaches the frontend from exactly one place. `acorn-daemon/src/server.rs` emits a `StreamFrame::Exit` when a session's broadcast channel reports `Closed`, and `daemon_stream.rs` re-emits that as the Tauri event `Terminal.tsx` listens for. A `tokio::sync::broadcast` closes only when its last `Sender` drops.

That sender lives on `PtyHandle`. Once `wait_loop` removes the handle from the live map, the remaining owner is the reader thread — and it holds its `Arc` for as long as it is inside `read()`. On Unix the reader wakes on EOF the moment the child is reaped, drops its `Arc`, and the close propagates. ConPTY does not close its read pipe when the child exits, so on Windows the reader stays parked, the sender is never dropped, and every attached client waits for output that will never arrive. `read_loop` samples its `stop` flag only *before* the blocking read, so tripping that flag from `wait_loop` does not wake it either.

## The change

`output_tx` becomes an `Option` that the wait thread clears once the child is reaped, immediately after the exit code is published so a pump woken by `Closed` still reports the real code. Closing no longer depends on the reader thread waking up. `subscribe` hands back an already-closed receiver if it races that clear, so a late attach gets an exit frame instead of parking on a dead channel.

Known ceiling, deliberately out of scope: the reader thread still parks until the PTY handle goes away, so Windows leaks one reader thread per exited session. Unblocking it means closing the master from another thread, which deserves its own change.

## Verification

- Reproduced on real hardware first: Windows 11, Acorn **v1.37.1** installed from the release NSIS. `exit` in a session left the shell dead (`powershell.exe` gone) while the UI stayed on `status: ready`, with no banner and no restart on Enter. An instant session and a project-backed session behaved identically.
- `subscriber_sees_broadcast_close_when_child_exits` covers the whole chain: spawn a child that exits on its own, deliberately hold an `Arc<PtyHandle>` across its exit to reproduce the parked-reader condition on any platform, then assert an attached subscriber observes the close. **Green on Windows CI**, and it fails on macOS when the one-line clear in `wait_loop` is reverted — so it guards the fix on both.
- The test answers the ConPTY cursor-position query (`ESC [ 6 n`). Without that the spawned shell blocks before its body and never exits, which is indistinguishable from a missed exit signal — the same trap the existing `powershell_conpty_accepts_input_resize_and_captures_output` test documents.
- `cargo test -p acorn-local-ipc -p acorn-platform -p acorn-daemon` (the set CI compiles) green locally; `cargo fmt --check` clean.
- Not re-verified end to end in the installed app on Windows: that needs an installer built from this branch, and CI does not publish the Windows sidecar as an artifact. The unit test covers the exact mechanism the app depends on.

Independent of #932, which is about verbatim `\\?\` paths in session create/remove. Different crates, no overlap.


im-ian:
Draft — do not merge. The Windows test in this branch is red on purpose: it pins a defect the fix here does **not** yet clear.

Two CI experiments on `windows-2025`, both with a `cmd /C ping -n 2 127.0.0.1` child that takes ~1s, sampled 15s after spawn:

| wait strategy in `wait_loop` | result |
| --- | --- |
| `child.wait()` (blocking, original) | `exit_code=None sender_cleared_after=None handle_arcs=4 last_recv=Empty` |
| `child.try_wait()` (polled, 100ms) | `exit_code=None sender_cleared_after=None handle_arcs=4 last_recv=Empty` |

`exit_code` is written as the first statement after the wait returns, so `None` means `wait_loop` never got past it. `handle_arcs=4` accounts for the handle map, the reader thread, the wait thread, and the test — both worker threads still alive. `last_recv=Empty` rules out a slow drain of the broadcast.

So neither `Child::wait()` nor `Child::try_wait()` in `portable_pty` ever reports the reaped ConPTY child on Windows. Everything the daemon publishes about exit is gated on that call — the exit code, the handle-map removal, the registry detach, and the broadcast close that `server.rs` turns into `StreamFrame::Exit` and `daemon_stream.rs` re-emits as `pty:exit:<id>`. That is why the app shows a dead shell as `Ready` with no restart prompt.

Clearing this needs an exit detector that does not go through `portable_pty`'s child status — an OS-level check on the pid (`GetExitCodeProcess` / waiting the process handle directly), or a `portable_pty` fix upstream. That is a daemon lifecycle change worth designing deliberately rather than bolting on.

The `output_tx: Option` half of this branch is independently correct and still required: even with a working exit detector, the broadcast would not close, because the reader thread's `Arc<PtyHandle>` keeps the sender alive — ConPTY never EOFs the read pipe and `read_loop` samples its `stop` flag only before the blocking read. Reverting just that half fails the test on macOS, so it is pinned.

im-ian:
Correcting my earlier comment on this PR — two of the conclusions in it were wrong, and the CI evidence that produced them was an artifact of a broken test fixture.

The instrumented runs reported `exit_code=None` because the spawned child **never exited**, not because `Child::wait()` failed to observe it. `scrollback_bytes=4` was the tell: `ESC [ 6 n` is exactly four bytes. Both `cmd.exe` and `powershell.exe` emit that cursor-position query during ConPTY startup and block until the terminal answers, so the child never reached its body. A non-PTY control spawn of the same command returned `Ok(Some(0))` in the same run, which should have pointed at the fixture rather than at `wait()`.

So, withdrawn:

- **`Child::wait()` is broken on Windows** — it is not. The `try_wait()` polling I pushed on the strength of that claim has been reverted; the blocking `wait()` is unchanged from `main`.
- **This needs an OS-level pid watchdog / a `portable_pty` fix** — it does not.

What stands, and is now verified on Windows CI rather than inferred: the broadcast does not close on its own, because the reader thread's `Arc<PtyHandle>` keeps the sender alive while it sits in a `read()` that ConPTY never ends. Clearing the sender in the wait thread is the fix, and the test — which holds an `Arc` across the child's exit on purpose to reproduce that condition anywhere — is green on Windows and red on macOS with the fix reverted.

